PR #53, opened 2013-03-02
Option to use vectorized versions of density callables

ipashchenko:
I found that using subject is quite easy thanks to numpy broadcasting and emcee nice code:) It significantly speeds up sampling and costs only tiny modification of callable density code (if numpy broadcasting is possible). 

Thanks for nice toolkit,
Ilya


dfm:
This is finally supported in the upcoming 3.0 release!

PR #62, opened 2013-04-22
Added progessbar to run_mcmc()

pirg:
Runs a progressbar for run_mcmc() using 
https://github.com/olivierverdier/python-progressbar 
if you have it installed.


dfm:
Something like this has been added to the upcoming 3.0 release using tqdm.

PR #99, opened 2014-02-09
Avoid RuntimeWarning when subtracting infs

mrgeorge:
In the _propose_stretch method of EnsembleSampler, occasionally both the new and proposed lnprob values are out of bounds of some prior and set to -np.inf. In cases where newlnprob and lnprob0 are both -np.inf (and type float64),

accept = A + newlnprob - lnprob0 > B

raises a warning for invalid value in subtract. The subtraction evaluates to nan, and accept is set to False.
On the other hand,

accept = A + newlnprob > lnprob0 + B

correctly evaluates to False with no warning. This commit makes that minor manipulation to avoid that warning message.

### DIFF
--- a/emcee/ensemble.py
+++ b/emcee/ensemble.py
@@ -332,8 +332,8 @@ class EnsembleSampler(Sampler):
         newlnprob, blob = self._get_lnprob(q)
 
         # Decide whether or not the proposals should be accepted.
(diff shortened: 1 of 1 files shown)


kinsey40:
My code was working fine, but always outputted the stated error. Above solution worked a treat. 

Thanks!

PR #100, opened 2014-02-10
Update ensemble.py to check initial state (Issue #99)

ipashchenko:
Added method `EnsembleSampler.check_state` that check given state for Nans & (optionally) Infs. Extended inherited from `Sampler` class method `run_mcmc` to check initial state for NaNs and (optionally) Infs. 
Edit: Actually, i don't think that PR is needed. I cited it here to illustrate Issue #99.


dfm:
This has all been checked for a while now. Sorry I never closed this.

PR #107, opened 2014-03-07
Move _function_wrapper up to Sampler class

ipashchenko:
Should it be in the base class? It saves time when one is subclassing `Sampler`.


dfm:
This isn't needed anymore because the base class no longer exists :-)

PR #129, opened 2014-08-10
Added reset_burn_in

### LINKED ISSUE #128: Add a reset_burn_in method
See the discussion here: #124

eteq:
This does what  #128 suggests and adds a `reset_burn_in` method.  So I think this closes #128

@dfm, I confess I don't understand why this works: I'm just reproducing the "old" `reset` prior to #124, but I don't fully understand why setting `iterations` and `naccepted` to 0 resets the chain... But I assume that's the right way since it worked before? 

cc @ipashchenko

### DIFF
--- a/emcee/sampler.py
+++ b/emcee/sampler.py
@@ -124,6 +124,15 @@ class Sampler(object):
         self.naccepted = 0
         self._last_run_mcmc_result = None
 
+    def reset_burn_in(self):
+        """
+        Clear the chain, while remember the last result from ``run_mcmc``.  This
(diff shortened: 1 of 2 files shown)


dfm:
I'm going to close this without merging because, in the 3.0 release, I'll recommend that people use the `discard` keyword in `get_chain`, etc. instead of the `reset` method because then you don't have to decide the length of your burn-in in advance. Feel free to re-open if you're not happy with that.

eteq:
That's exactly what I've been doing on my own most of the time anyway, so sounds great to me!

PR #131, opened 2014-09-16
New example: multimodal.py, to test PTSampler evidence estimation

drphilmarshall:
It's just the example in the docs, plus a little evidence check. Here's the result:

```
Estimated log evidence =  -16.9586413385 +/- 7.29507626509
Analytic log evidence =  -6.67931612501
```

which means that there is at least one bug in this code! But it runs to completion, and the docs are up to date. What more could one want?

### DIFF
--- a/docs/user/pt.rst
+++ b/docs/user/pt.rst
@@ -123,7 +123,16 @@ The resulting samples (1000 of them) are stored as the
     # Longest autocorrelation length (over any temperature)
     max_acl = np.max(sampler.acor)
 
-    # etc
+    # log evidence estimation
+    lnZ, dlnZ = sampler.thermodynamic_integration_log_evidence()
+
+    # Analytic log evidence, for comparison
+    lnZtrue = np.log(2.0*np.pi*0.1 * 2.0)
+
+
+The code for this example is available 
+`here
+<https://github.com/dfm/emcee/blob/master/examples/multimodal.py>`_.
 
 
 Implementation Notes
--- a/emcee/ptsampler.py
+++ b/emcee/ptsampler.py
@@ -92,6 +92,9 @@ class PTLikePrior(object):
 
         if lp == float('-inf'):
             return lp, lp
+        elif np.isinf(lp):
+            lp = float('-inf')
+            return lp, lp
 
         return self.logl(x, *self.loglargs, **self.loglkwargs), lp
 
--- a/examples/multimodal.py
+++ b/examples/multimodal.py
@@ -1,0 +1,103 @@
+import numpy as np
+from emcee import PTSampler
+
+# Example from PTsampler docs at http://dan.iel.fm/emcee/current/user/pt
+# 2 well-separated Gaussians, evidence is known analytically.
+
+# Define the means and standard deviations of our multi-modal likelihood:
+
+# mu1 = [1, 1], mu2 = [-1, -1]
+mu1 = np.ones(2)
+mu2 = -np.ones(2)
+
+# Width of 0.1 in each dimension, circularly symmetric:
+sigma = 0.1
+ivar = 1.0/(sigma*sigma)
+sigma1inv = np.diag([ivar, ivar])
+sigma2inv = np.diag([ivar, ivar])
+
+def logl(x):
+    dx1 = x - mu1
+    dx2 = x - mu2
+    return np.logaddexp(-np.dot(dx1, np.dot(sigma1inv, dx1))/2.0,
+                        -np.dot(dx2, np.dot(sigma2inv, dx2))/2.0)
+
+# Use a 2D uniform prior, correctly normalized:
+def logp(x):
+    xmax = 5.0
+    if (abs(x).any() > xmax): 
+        logp = -np.inf
+    else:
+        logp = -2.0*np.log(2.0*xmax)
+    return logp
+
+# (Approximate) analytic evidence for two identical Gaussian blobs,
+# over a uniform prior [-5:5][-5:5] with density 1/100 in this domain:
+
+log_evidence = np.log(2.0 * 2.0*np.pi*sigma*sigma / 100.0)
+
+# Now we can construct a sampler object that will drive the PTMCMC; arbitrarily,
+# we choose to use 20 temperatures (the default is for each temperature to
+# increase by a factor of sqrt(2), so the highest temperature will be T=1024,
+# resulting in an effective sigmaT=32sigma=3.2, which is about the separation of
+# our modes). Let's use 100 walkers in the ensemble at each temperature:
+
+ntemps = 20
+nwalkers = 100
+ndim = 2
+
+sampler = PTSampler(ntemps, nwalkers, ndim, logl, logp)
+
+# Making the sampling multi-threaded is as simple as adding the threads=Nthreads
+# argument to PTSampler. We could have modified the temperature ladder using the
+# betas optional argument (which should be an array of beta = 1/T values). The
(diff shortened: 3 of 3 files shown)


farr:
Hi Phil, David, all,

Sorry for taking so very loooooooooong to get back to this.  I think I understand what the problem is, and it's not a bug in the code.  Instead, it's a bug in the "test."  The example from the docs has a flat, improper prior (log_prior = 0.0 everywhere), so the TI integral does not converge:

\int d\beta <\log L>_\beta = \int d\beta -1/(2 \beta)

is, in fact, infinite---it diverges like log(x) as x --> 0.  The evidence is well-defined---it is just the integration method that fails.

Embarrassingly I was bitten by this same bug when writing the PT Sampler tests (note that there is---or at least was---an evidence check in the test() routine for PT Sampler), but didn't recognise it again here.  The solution would be to put in a prior that is wide (maybe flat between +/- 10-sigma?), and ensure that the highest temperature ladder can sample from it.

ipashchenko:
@farr Your suggestion explains my correct result with ``PTSampler`` that i cited here above - i have used proper prior. I think that issue with evidence estimation must be explained in docs or even print a warning message when calling ``PTSampler.thermodynamic_integration_log_evidence``:)

farr:
I've added the suggested changes to the docs in #154 ; Dan, if you pull
those changes into emcee, you can close this issue, too.

drphilmarshall:
An improper prior error! That's a rare treat :-) What does the
multimodal.py test return now, for the evidence values?

farr:
I haven't tried it on the multimodal.py, but it passes with flying colours
the evidence computation and sampling on a unimodal Gaussian in
check_pt_sampler (line 123 of tests.py).  If you want to give it a go, you
can have a look at the code there---it took me forever to pass that test
because I originally used an improper prior, and couldn't understand why I
was failing.  As I said, it's a bit embarrassing that I forgot all about
that experience when thinking about your bug report.

drphilmarshall:
Hmm. `examples/multimodal.py` does already have a proper prior, it turns out:

```python
# Use a 2D uniform prior, correctly normalized:
def logp(x):
    xmax = 5.0
    if (abs(x).any() > xmax): 
        logp = -np.inf
    else:
        logp = -2.0*np.log(2.0*xmax)
    return logp
```

What could be wrong with this?

farr:
I think the line 

    if (abs(x).any() > xmax):

is a mistake.  It should be written

    if np.any(np.abs(x) > xmax):

For example:

    x = array([1.0, 2.0])
    if x.any() > 1:
        print 'Yep'
    else:
        print 'Nope'

prints 'Nope', while 

    if np.any(abs(x) > 1):
        print 'Yep'
    else:
        print 'Nope'

prints 'Yep'.  So, the prior is, in fact, unnormalised.  I fixed that line, and increased the temperature density in multimodal.py (after examining a plot of the TI integrand versus beta, and seeing a lot of jagged structure), and now it works:

    Estimated log evidence =  -6.89598100431 +/- 0.0546850237424
     Analytic log evidence =  -6.67931612501

These changes are pushed to the tip of the fix-pt-docs branch referred to in pull request #154, so pulling that branch should also resolve this issue.

ipashchenko:
@farr Will, could you please add some potential diagnostics tools for better temperature settings somewhere in docs? You wrote that you have examined a plot of the TI integrand versus beta... As far as i understand there's no universal temperature setting.

farr:
@ipashchenko Eventually, yes.  That's a bit of a long-term project.  

You're right that there is no universal temperature setting.  One attempt to address this issue for sampling is http://arxiv.org/abs/1501.05823 .  For TI, just looking at the plot of the integrand is usually enough to diagnose problems, but the uncertainty reported by the TI method can help, too (i.e. if it's large, then you probably need a denser T spacing).

Anyway, this is a bit like universal methods for determining burnin: there are none, and the topic is very broad.

drphilmarshall:
Great, thanks Will! Sorry about the crap coding - although I am glad that
my bug was in use of numpy rather than understanding of how to apply
probability theory.

farr:
No worries re: coding---I'm just glad to have (finally!) tracked down and
resolved the issue!

drphilmarshall:
Yeah! Also: fantastic to see evidence values coming out of the PTSampler!
:-)
@kbarbary, your dreams are being fulfilled here.

drphilmarshall:
OK, something's weird here: I pulled down @farr's fix-pt-docs branch into a fix-pt-docs branch of my own, and tried to make `multimodal.py` work. I get:
```
Estimated log evidence =  7.26355058317 +/- 1.32035266078
Estimated log evidence =  7.26947875152 +/- 1.3040628331
Analytic log evidence =  -6.67931612501
```
in two runs of the code. The `np.any()` lines is there, and I'm using @farr's cooling schedule...

farr:
There was some confusion on my part with a "-" sign in the integral earlier
today when I transitioned to using trapz. Can you check that you have
commit 81782de? If you do, I'm not sure what is going on....

drphilmarshall:
Yep, I've got that commit. Here's [my branch](https://github.com/drphilmarshall/emcee/tree/fix-pt-docs) if you want to check it out.

farr:
Hmm.  Here's a transcript of my session:

    eAmu:emcee farr$ git checkout marshall/fix-pt-docs
    # A bunch of warnings about detached head.

    eAmu:emcee farr$ python setup.py install
    # A bunch of install crap
 
    eAmu:emcee farr$ cd examples/
    eAmu:examples farr$ ipython --pylab

    In [1]: %run multimodal.py
    PT burning in for 100 iterations...
    PT sampling for 1000 iterations...
    Temperature swap acceptance rates are 
    T =  1000000.0  accept =  0.99983
    T =  621016.941892  accept =  0.999745
    T =  385662.042116  accept =  0.99966
    T =  239502.661999  accept =  0.99949
    T =  148735.210729  accept =  0.999145
    T =  92367.0857187  accept =  0.998615
    T =  57361.5251045  accept =  0.9978
    T =  35622.4789026  accept =  0.996665
    T =  22122.1629107  accept =  0.99468
    T =  13738.2379588  accept =  0.990925
    T =  8531.67852417  accept =  0.985705
    T =  5298.31690628  accept =  0.977805
    T =  3290.34456231  accept =  0.96518
    T =  2043.35971786  accept =  0.94696
    T =  1268.96100317  accept =  0.922105
    T =  788.046281567  accept =  0.88864
    T =  489.390091848  accept =  0.85261
    T =  303.919538231  accept =  0.821315
    T =  188.739182214  accept =  0.801945
    T =  117.210229753  accept =  0.79432
    T =  72.7895384398  accept =  0.78735
    T =  45.2035365636  accept =  0.780655
    T =  28.0721620394  accept =  0.772305
    T =  17.433288222  accept =  0.765715
    T =  10.8263673387  accept =  0.765155
    T =  6.7233575365  accept =  0.766605
    T =  4.17531893656  accept =  0.76775
    T =  2.5929437974  accept =  0.766465
    T =  1.61026202756  accept =  0.76507
    T =  1.0  accept =  0.76491

    Estimated log evidence =  -6.87513567321 +/- 0.085128437244
     Analytic log evidence =  -6.67931612501

Did yours print all the temperature and swap acceptance info?  If not, could it be that you were running of an older commit because you didn't install?  I'm sort of at a loss here to explain the different behaviour....

drphilmarshall:
Beats me. Feel free to close this, and I'll just pull from the base repo
master branch once it all gets merged in properly. Thanks!

PR #138, opened 2014-11-02
make the quickstart example work on Python3

stonebig:


dfm:
Sorry for taking infinity years to respond to this, but I'm in the process of re-writing all the examples finally and I'll take this into account in the next release. Thanks!

PR #139, opened 2014-11-18
Closing properly the pool in the Sample function ath the end of the run.

jeammimi:
The pool is closed only if it was created by the emcee module.

When runing a sampler with threads different from 1 in a ipython notebook,
the pool was not closed properly and everytime I was using the sampler, a
new pool was added and not close.
This close the pool at the end of the run.


dfm:
This is great, but the upcoming 3.0 release will deal with this problem by just getting rid of the "threads" argument - that simplifies a lot of issues! Thanks for your contribution.

PR #146, opened 2015-01-29
Adaptive parallel tempering

willvousden:
Dynamic temperature selection according to [arXiv:1501.05823](http://arxiv.org/abs/1501.05823v1.pdf).


dfm:
I'm going to close this because in v3.0 the PTSampler has been removed and a shim to [willvousden/ptemcee](https://github.com/willvousden/ptemcee) has been added.

carlosgmartin:
@dfm The ptemcee repository seems to be abandoned and lacks documentation. Are there any plans to incorporate it (the adaptive parallel tempering algorithm) into emcee, or is there any current way to use it in emcee?

dfm:
This will not be incorporated into emcee. If you need something like PT and premcee doesn't cut it, I'd recommend [dynesty](https://dynesty.readthedocs.io/en/latest/) instead!

PR #152, opened 2015-05-14
Move import of mpi4py to inside MPIPool initialization

joezuntz:
We have a downstream user with a broken MPI installation, which crashes on importing mpi4py. She also can't use emcee in serial mode, because the mpi_pool module tries importing mpi4py.MPI.

This change fixes that (certainly rare) issue by moving the mpi4py import to inside the MPIPool initialization.


marius311:
I don't think this is that rare. Another example is often times if you're on the login node of a cluster, anything touching MPI will error out, so you can't test emcee unless you get a compute node. This PR fixes the issue for me in the couple of places I've run into it, it'd be great to have it merged.

dfm:
Thanks! Sorry about the delay.

PR #154, opened 2015-05-22
Fixes #131, issues with thermodynamic integration evidence

farr:
Updates to the documentation to warn against using improper priors when computing evidence using TI.  This is the issue encountered in #131.

### DIFF
--- a/emcee/ptsampler.py
+++ b/emcee/ptsampler.py
@@ -4,7 +4,7 @@
 from __future__ import (division, print_function, absolute_import,
                         unicode_literals)
 
-__all__ = ["PTSampler"]
+__all__ = ["PTSampler", 'default_beta_ladder']
 
 import numpy as np
 import numpy.random as nr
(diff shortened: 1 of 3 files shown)


emcee/ptsampler.py
```diff
@@ -465,16 +468,16 @@ def thermodynamic_integration_log_evidence(self, logls=None, fburnin=0.1):
...
+            betas2 = self.betas[::2]
 
             istart = int(logls.shape[2] * fburnin + 0.5)
 
             mean_logls = np.mean(np.mean(logls, axis=1)[:, istart:], axis=1)
             mean_logls2 = mean_logls[::2]
 
-            lnZ = -np.dot(mean_logls, np.diff(betas))
-            lnZ2 = -np.dot(mean_logls2, np.diff(betas2))
+            lnZ = np.trapz(mean_logls, betas)
```
ga7g08: Are you not missing the `-` here?

farr: I checked it in the tests, and the "-" is inappropriate here.

ga7g08: I checked your branch out locally and ran the tests and they failed. The issue is (I'm almost certain) those `-`'s, the integral is still from 1 to 0 and so needs the minus.

farr:
OK @ga7g08, does this fix your issues with signs, etc?  It passes all my tests.  I just enforced an increasing ordering of the betas when doing the TI.

ga7g08:
Yep, all works fine for my examples. This is nice as it makes it very readable.

abeelen:
Hi,

It seems that this PR has not been merged yet, any particular reason for that ? Looks to me like this alone should drive a new release.

ga7g08:
Hi, just bumping this as I forgot about it, started using master, and came into the same problem: systematic errors in the thermodynamic_integration_log_evidence calculation. Is there a reason not to merge this?

pkienzle:
Shouldn't ```betas2``` and ```mean_logls2``` be derived from the sorted betas?  

This can be done using
```
    lnZ2 = np.trapz(mean_logls[::2], betas[::2])
```

dfm:
I'm going to close this because in v3.0 the PTSampler has been removed and a shim to [willvousden/ptemcee](https://github.com/willvousden/ptemcee) has been added. I assume that this is done correctly in ptemcee but, if not, let's open this over there!

ljschumacher:
looking at the code https://github.com/willvousden/ptemcee/blob/master/ptemcee/util.py
ptemcee does seem to use the improved integration method described here

PR #155, opened 2015-05-22
Improve the td integration method in ptsampler

ga7g08:
Previously this used a simple quadrature, but this introduced a bug
which tended to end in overestimates of the evidence. Replacing this
with numpy's trapz rule avoid this bug and improves the readability

@farr : Could you take a look at this and check I haven't misunderstood
the original code?

### DIFF
--- a/emcee/ptsampler.py
+++ b/emcee/ptsampler.py
@@ -465,16 +465,16 @@ class PTSampler(Sampler):
             return self.thermodynamic_integration_log_evidence(
                 logls=self.lnlikelihood, fburnin=fburnin)
         else:
-            betas = np.concatenate((self.betas, np.array([0])))
-            betas2 = np.concatenate((self.betas[::2], np.array([0])))
+            betas = self.betas
+            betas2 = self.betas[::2]
 
             istart = int(logls.shape[2] * fburnin + 0.5)
 
             mean_logls = np.mean(np.mean(logls, axis=1)[:, istart:], axis=1)
             mean_logls2 = mean_logls[::2]
 
-            lnZ = -np.dot(mean_logls, np.diff(betas))
-            lnZ2 = -np.dot(mean_logls2, np.diff(betas2))
+            lnZ = np.trapz(mean_logls, betas)
+            lnZ2 = np.trapz(mean_logls2, betas2)
 
             return lnZ, np.abs(lnZ - lnZ2)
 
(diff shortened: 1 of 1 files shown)


farr:
I have a couple of questions:

* Are you sure that your method is more accurate?  It will be if one of the
betas in the ladder is 0, but if you don't include a T = infinity, beta = 0
chain, then you will find that it is systematically off, since it ignores
the area between min(betas) and beta = 0.  That was the rationale behind
the original Riemann sum.

* What happened to the "-"?  You want to integrate from beta = 0 to beta =
1, but the betas are stored in the opposite order in the array, right?

ga7g08:
Firstly I am extremely embarrassed to admit that I didn't run the test properly before
pushing (I forgot to install the modified version before running the test) and hence 
dropped the minus. I've added that in now and perhaps before this is ever pulled in
I will  squash it into one sensible commit. 

Secondly thanks for explaining the rationale for the Riemann sum, I hadn't appreciated that
that was the intention. However, since numpy calculated the difference as `out[n] = a[n+1] - a[n]`, then the implementation is a right-Riemann sum and therefore is an overestimation since <loglike> is monotonically increasing. 

So it seems my method neglects the beta -> 0 part of the integration, while the right-Riemann sum over estimates the integral. Personally I feel that the user should be checking (or perhaps should be forced to check) the plot of beta againat lnlike to ensure they have reasonable values (e.g lnlike is sufficiently small for there smallest beta). In this case I prefer trapz since it is more readable, but the obvious alternative is a middle Riemann sum?

PS: I am wondering if the issue discussed in #131 may be solved here? I ran the `multimodal.py` example as given in @drphilmarshall commit and got 

```
Estimated log evidence =  -64.0249172494 +/- 152.159339428
Analytic log evidence =  -6.67931612501
```

PPS: If you want an example of the overestimation [here is a gist containing the minimal code](https://gist.github.com/ga7g08/2874c19a55f0a2452135) with results:

```
Actual ln(Z) = 93.4636629474
lnZ_m1=93.1525160422, err=0.311146905153
lnZ_m2=95.5127068853, err=2.04904393787
```

farr:
Actually, there is a bug in Marshall's multimodal.py, which skips the
normalisation branch of the if statement in the prior.  I uploaded a fixed
version, and transitioned to your trapz integration (you're right---it's
better to have a more accurate integral over the range than to try to
squeeze the last bit of integrand toward beta = 0) in PR #154.  Go ahead
and have a look.

ga7g08:
Okay thanks, I added a note that I think you are missing a minus sign as I did ( I guess it has been a long day :) ) Should I close this then as it is no longer needed?

farr:
I think you can go ahead and close it, once you verify that #154 works for
you as it did for me.

PR #160, opened 2015-08-08
Fix PTSampler.run_mcmc()

pkgw:
The generic Sampler.run_mcmc() function wouldn't work right for PTSampler
because it passed the RNG state, `rstate0`, as the second positional argument
to sample(), while for PTSampler the second positional argument was the extra
`lnlike0` parameter. If you tried to pass `lnlike0` yourself, then, you'd
get an exception for a doubly-specified keyword.

Anyway, the best solution is probably to pass `rstate0` as a keyword argument so
that subclasses can have differently-ordered arguments if they want. Do that.
This also requires fixing a discrepancy between the documentation and
implemention of the M-H sampler, which claimed the RNG state keyword was
`rstate0` when it was really being called `randomstate`.

### DIFF
--- a/emcee/mh.py
+++ b/emcee/mh.py
@@ -50,7 +50,7 @@ class MHSampler(Sampler):
         self._chain = np.empty((0, self.dim))
         self._lnprob = np.empty(0)
 
-    def sample(self, p0, lnprob=None, randomstate=None, thin=1,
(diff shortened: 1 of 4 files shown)


pkgw:
It's been a year, but this is still mergeable ...

dfm:
Also this. Thanks again!

PR #161, opened 2015-08-11
New example and doc use of thermodynamic integration

ga7g08:
This adds a new example `pt_thermodynamic_integration.py` to illustrate
the use of the thermodynamic integration method. A section is also added
to the documentation to help familiarise users with the process.

@farr I've added what I think is the simplest thing: an example script and an additional section in the docs. The description at the end of the docs is sadly lacking detail as I am still confused about the subject myself! Let me know what you think.


ga7g08:
I've just realised the axes label should be Mean log-likelihood. I'll change this.

docs/user/pt.rst
```diff
@@ -187,3 +187,85 @@ normalized), we have
...
+    sampler = emcee.PTSampler(ntemps, nwalkers, ndim, logl=logl, logp=logp,
+                              betas=betas)
+    p0 = np.random.uniform(-1, 1, size=(ntemps, nwalkers, ndim))
+    out = sampler.run_mcmc(p0, 200)
+
+Then finally we get the evidence and an estimate of the error::
+
+    ln_evidence, ln_error = sampler.thermodynamic_integration_log_evidence()
+    evidence = np.exp(ln_evidence)
+    error = abs(ln_evidence) * ln_error
```
bamford: This should be
```python
error = evidence * ln_error
```
but is correct in code (see below).

examples/pt_thermodynamic_integration.py
```diff
@@ -0,0 +1,39 @@
...
+ndim = 1
+betas = np.logspace(0, -3, ntemps)
+sampler = emcee.PTSampler(ntemps, nwalkers, ndim, logl=logl, logp=logp,
+                          betas=betas)
+p0 = np.random.uniform(-1, 1, size=(ntemps, nwalkers, ndim))
+out = sampler.run_mcmc(p0, 200)
+
+ln_evidence, ln_error = sampler.thermodynamic_integration_log_evidence()
+evidence = np.exp(ln_evidence)
+error = np.exp(ln_evidence) * ln_error
```
bamford: This could simply be
```python
error = evidence * ln_error
```

ga7g08:
Thanks @bamford  for spotting those errors.

dfm:
I'm going to close this because in v3.0 the PTSampler has been removed and a shim to [willvousden/ptemcee](https://github.com/willvousden/ptemcee) has been added.

Issue #163, opened 2015-08-26
Progress bar while sampling


tomevans:
Is it possible to display a progress bar while sampling, to show what fraction of the total steps have been taken as a function of time? e.g. pymc has a nice routine for doing this.

ipashchenko:
Hi!
Please, see PR #62 concerning progress bar addition and ``emcee`` author's  https://github.com/dfm/emcee/issues/133#issuecomment-57087428 concerning difference between ``emcee`` and ``pymc``.

tomevans:
Thanks for the reply, Ilya. If I understand correctly then, it should be possible to wrap a progress bar routine *around* a run_mcmc() call? i.e. without having to incorporate a progress bar into the emcee code itself... If anyone has a snippet of code for doing this, I'd appreciate seeing it, as I couldn't work it out and didn't see it in the docs.

dfm:
You can follow a pattern [like this](http://dan.iel.fm/emcee/current/user/advanced/#incrementally-saving-progress). For the simplest progress bar, this would look something like:

```
nsteps = 5000
for i, (pos, lnp, state) in enumerate(sampler.sample(p0, iterations=nsteps)):
    if (i+1) % 100 == 0:
        print("{0:.1f}%".format(float(i) / nsteps))
```

In practice you could use something like the [progressbar](https://pypi.python.org/pypi/progressbar) module but it will look very much like this example.

I'd be happy to add an example to the docs if someone submitted it ;-)

willvousden:
I've used [clint](https://github.com/kennethreitz/clint) for this in the past.  I'll hack together an example.

ipashchenko:
Hi, Will!
Could you please add Dan's example implementing progress bar without third party packages (https://github.com/dfm/emcee/issues/163#issuecomment-135460483) to your PR?

willvousden:
Sure thing.  I've replaced the clint example with Dan's example and another to illustrate how to make an actual progress bar.

tomevans:
Thank you all - I've finally gotten around to implementing it myself, following your suggestions.

GaelO:
Hi ! I tried looking around for a solution to my problem but didn't find any: 
I implemented a progress bar to the **Fitting a Model to Data** example. But it appears to add an _nsteps_  MCMC, and then launch a second one right behind with the _N_ steps argument from _run_mcmc_.
Here's a example of such a chain, with _nsteps = 300_, and _N=500_

[...]

PR #164, opened 2015-09-02
Updated docs with progress bar example.

willvousden:
Added progress bar example (for clint) to docs in response to #163.


dfm:
Thanks!

PR #170, opened 2015-10-19
Update the dependency on triangle module to corner

migueldvb:


dfm:
Thanks! :beers:

davidwhogg:
I love that it requires an external pull request to keep two of your *own codes* compatible!  :)

PR #172, opened 2015-12-08
emcee page affiliation update

eteq:
OK, this is definitely nitpicky... But when I realized this page's source was on github I figured I can just do this to update my affiliation ;)


dfm:
Might be better to remove affiliations… :-)

eteq:
And waste all my effort on this PR? (all ~20 seconds) Bah!

PR #178, opened 2016-05-18
CI: Test on python 3.4 and 3.5. Add coverage.

ericdill:
Commit 1 adds testing for python 3.4 and 3.5 to travis-ci.

Commit 2 collects coverage stats for the test suite, uses codecov to parse and
render the output and adds a badge from codecov to the readme.  Happy to remove
the second commit if you would like.


ericdill:
Oh, I can also add nightly python testing if you want, though that seems slight overkill

ericdill:
Apparently the coverage module is not python 3.2 compatible.  Would you like to

1. Don't gather coverage stats and thus don't need to worry about this error
2. Skip gathering of coverage stats for python 3.2 only
3. Drop python 3.2 support from CI

dfm:
I'd be happy to drop Py3.2. Thanks for this!

ericdill:
bump @dfm

dfm:
Thanks!

PR #180, opened 2016-05-21
Installation section added to readme file

Volodymyrk:
just added `pip install emcee` to readme file


dfm:
I'd rather just point users to the docs as it already is because then they can see all the options. Thanks!

PR #192, opened 2016-08-24
Emcee demc

mdanthony17:
This branch adds the differential evolution MCMC algorithm described in the papers by Ter Braak and Nelson, Ford, and Payne.  Required inputs are very similar to ensemble sampler however algorithm proposes steps and accepts them differently.  Performance is comparable to affine invariant sampler in most distributions but gives drastic improvements when the posterior is multimodal.


dfm:
Hi. Thanks for this contribution. I'm actually not going to merge this because I'm pushing all new features to [emcee3](https://github.com/dfm/emcee3) where they will be easier to maintain and I already implemented [DE](https://github.com/dfm/emcee3/blob/master/emcee3/moves/de.py). Given how much simpler this implementation is, I think you'll appreciate why I want to keep it there, right?

Thanks again and sorry for the slow response!

PR #204, opened 2016-11-24
Improve testing and setup.py

aragilar:
This adds support for using tox, which makes it easy to test against different python versions. It also cleans up the setup.py and add support for making universal wheels.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ docs.tar
 *.pdf
 
 .coverage
+
+.tox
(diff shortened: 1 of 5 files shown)


dfm:
Thanks for the contribution. I'm pretty happy with the travis setup as it is and I've had nothing but headaches from the previous times that I've accepted pull requests implementing tox support. So, unless there's a huge outpouring of support from the community to accept this, I'm not inclined to merge this full pull request. I'd be happy to consider the change to `setup.py` if you want to submit a separate pull request with just that change. Thanks again!

PR #206, opened 2016-12-03
Clean up the setup.py and add support for making universal wheels.

aragilar:
Replacement for #204 without tox and travis changes.


dfm:
I've incorporated this. Thanks!

PR #208, opened 2016-12-12
Vectorize lnprobfn call

nkern:
I added the capability to make lnprobfn() call the entire vector p all at once, instead of using map() to iterate over p. For custom lnprobfn() functions that are already highly vectorized, this can bring a large speed-up (for my custom use-case I get over an order of magnitude in speed-up).

### DIFF
--- a/emcee/ensemble.py
+++ b/emcee/ensemble.py
@@ -79,13 +79,13 @@ class EnsembleSampler(Sampler):
     """
     def __init__(self, nwalkers, dim, lnpostfn, a=2.0, args=[], kwargs={},
                  postargs=None, threads=1, pool=None, live_dangerously=False,
-                 runtime_sortingfn=None):
+                 runtime_sortingfn=None, vectorize=False):
         self.k = nwalkers
         self.a = a
         self.threads = threads
         self.pool = pool
         self.runtime_sortingfn = runtime_sortingfn
-        self.vectorize = False
+        self.vectorize = vectorize
 
         if postargs is not None:
             args = postargs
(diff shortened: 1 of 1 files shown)


nkern:
I should note that the edits are in ensemble.py, and the vectorized lnprobfn() is toggled on or off via the keyword-argument "vectorize" which is set to False by default.

ipashchenko:
Hi! I've implemented smth similar in PR#53. It really speeds up sampling sometimes.

dfm:
Cool. Can you have a go at adding a usage example, some notes in the docs (maybe in "advanced patterns"), and some tests for this?

- It should test that it performs properly under the correct conditions but it would be interesting to also see what happens if it is used incorrectly. Are useful exceptions thrown at the right times or does it just return garbage?
- What if the results returned by the lnprob function are the wrong shape?
- Does this support blob vectors too?
- Is "vectorize" the right name for this argument?

Thanks and let me know if you need a hand getting this set up.

manodeep:
If I understand correctly, `vectorize` passes off **all** points into the `lnprob` function simultaneously. Since the `loadbalance` (and `runtime_sorting`) option is implemented assuming single evaluations, the `vectorize` and the `loadbalance` options are likely to be mutually exclusive. If so, a simple check that both options are not enabled might be a good idea.

nkern:
Hi there,

@dfm: yes, I'll look into making some tests and checking compatibility with blob vectors. About the name of the on-off flag, I thought of "vectorize" because it is operating on the whole p vector (allowing custom-lnprobfn functions that are by design vectorized to be exploited) rather than iterating over the p vector with the MAP function. But it could also be something else. Let me know if you have any ideas.

@manodeep: I'll look into that, shouldn't be hard.

Here is a benchmark for my personal use-case (I've written personal functions that wrap around emcee's run_mcmc but they are in effect just calling run_mcmc). I think that in other cases vectorize=True does not lead to a speed up and may actually lead to producing more overhead. But I'll look into making a generic use-case.

screen shot 2016-12-12 at 6 54 00 pm

Best,
Nick

PR #213, opened 2017-03-01
Create .gitattributes to fix language statistics

roliveira:
The current language statistics are taking in account the `document`
folder. This shows the project as only ~43% Python... The `.gitattributes`
file fixes this by excluding that folder and additionally the `Makefile` in
the root folder as well.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+document/* linguist-documentation
+Makefile linguist-vendored=true


dfm:
Cool. I didn't know about this. Thanks!

roliveira:
It's a neat feature to [override the language definitions](https://github.com/github/linguist#overrides)! They usually do a go job ignoring certain folders, such as the `docs` one that is ignored by [default](https://github.com/github/linguist/blob/master/lib/linguist/documentation.yml). 

However... My fork did immediately updated the statistics, yours seems to be taking a little longer... 

But hey! Thanks for the package!!!

PR #220, opened 2017-07-07
Autocorrelation fix

farr:
Hi Dan,

I found a bug that resulted in always-failing multidimensional ACL estimates, even when successive single-dimensional estimates succeed.  Along the way, I noticed that your computation of the ACL is actually O(N^2) because you repeat O(N) computation for each value of the window function; I took the opportunity to fix that, too.

Will

### DIFF
--- a/emcee/autocorr.py
+++ b/emcee/autocorr.py
@@ -58,7 +58,7 @@ def integrated_time(x, low=10, high=None, step=1, c=10, full_output=False,
         low (Optional[int]): The minimum window size to test. (default: ``10``)
         high (Optional[int]): The maximum window size to test. (default:
             ``x.shape[axis] / (2*c)``)
-        step (Optional[int]): The step size for the window search. (default:
+        step (Optional[int]): The step size for the window search. (default: 
             ``1``)
         c (Optional[float]): The minimum number of autocorrelation times
             needed to trust the estimate. (default: ``10``)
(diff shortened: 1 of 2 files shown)


dfm:
I'm cool with this change but can you tell me what error is thrown for you? Your test already passes for me!

farr:
Huh. Originally I was getting "chain too short" errors even with ridiculous lengths of the chain. But maybe I was just unlucky. In any case, you will appreciate the O(N) scaling for chain lengths in the 1000s and no "high" parameter to narrow the window. 

Will

dfm:
Absolutely! Great find. I'm just confused/worried if this gives different results because it should give the same answer to machine precision.

farr:
Well, that's what I thought at first, but no.  The n-D version will look for a *common* window for all the dimensions that satisfies the constraints; doing n 1-D versions will allow the window to differ.

Will

dfm:
I agree and I'm planning on addressing that but your pull request doesn't change that fact does it? If it does, I don't see it.

dfm:
In other words, I don't think that this is a "fix" :-) Instead, I would say that it improves performance... still greatly appreciated but I want to make sure that I'm not missing something else!

farr:
Yeah, I went back at tried my testing on the nD version again, and it passes here, too.  I guess I was just using too short a sequence (though I think I tested at least into the thousands) when testing before.  Of course, the chance of failure is n times greater in nD than in 1D, and given the window issue it seems easy to have nD fail when iterated 1D succeeds; but that's still true in the new version, as well.

Sorry for the confusion---indeed, this is just a performance improvement, not a bug fix.

Will

dfm:
Excellent. Thanks!

I often do a per-parameter or some other custom autocorr calculation because of exactly this issue so I've been thinking about how to address it properly, but I haven't gotten around to implementing anything. Either way, this will help a lot!

farr:
Yeah, tbh I don't think there is a need to be so fancy with the nD calculation; just iterating the 1D calculation n times is at most a small constant factor overhead (and if the ACL is dominating your analysis, you're doing something wrong).  So maybe just rip it all out and do the 1D thing n times?

(That's what I do in my own package for this: https://github.com/farr/plotutils/blob/master/plotutils/autocorr.py ).

Will

PR #223, opened 2017-08-08
Adds exponential_time calculation to autocorr

ga7g08:
Currently, `autocorr` only provides the integrated autocorrelation time. This adds an estimate of the exponential autocorrelation time based on the method proposed in [http://www.sciencedirect.com/science/article/pii/S221313371300022X](http://www.sciencedirect.com/science/article/pii/S221313371300022Xl).

### DIFF
--- a/emcee/autocorr.py
+++ b/emcee/autocorr.py
@@ -125,6 +125,42 @@ def integrated_time(x, low=10, high=None, step=1, c=10, full_output=False,
                         "the autocorrelation time")
 
 
+def exponential_time(x, axis=0, fast=False):
+    """Estimate the exponential autocorrelation time of a time series.
+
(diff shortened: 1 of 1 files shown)


dfm:
Thanks for the contribution! I don't think that I can merge this because I've done a lot of experiments and I find that (despite what your reference says) this is pretty much never a good estimator for the integrated autocorrelation time. It's not that I don't think that a maximum likelihood estimator is a good idea, but I haven't found any problems where AR(1) is good! It's possible that an AR(2) estimator would be better, but I haven't really done the tests to confirm that. There is some discussion of some updates to the autocorrelation time estimation routine for the v3 release here: http://emcee.readthedocs.io/en/latest/tutorials/autocorr/

Thanks again and sorry for being slow.

PR #224, opened 2017-08-10
Fix typo in likelihood code block

benjaminrose:
Minus error in code block of the likelihood function. The math is correct but the code had an error.


dfm:
Close but no cigar. Double check and remember that the thing in the log is the *inverse* variance! :wink:

(You're not the first to make this mistake - myself included.)

benjaminrose:
So true. I looked several times assuming I was wrong. I focused mostly on parentheses and forgot about what `inv_sigma2` was.

PR #230, opened 2017-09-07
When threads>1, a finalizer calls InterruptiblePool.terminate when an…

lciti:
… EnsembleSampler object is deleted

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ python:
 install:
   - pip install numpy
   - pip install nose
+  - pip install "backports.weakref;python_version<'3.4'"
   - pip install coverage
   - pip install codecov
 
--- a/emcee/ensemble.py
+++ b/emcee/ensemble.py
@@ -14,6 +14,7 @@ from __future__ import (division, print_function, absolute_import,
 __all__ = ["EnsembleSampler"]
 
 import numpy as np
+import weakref
 
 from . import autocorr
 from .sampler import Sampler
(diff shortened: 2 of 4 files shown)


lciti:
This pull request addresses the fact that `EnsembleSampler` does not release the pool of workers when the object is destroyed. For example, the following adaptation of the quickstart example keeps creating new processes at each iteration without ever releasing the old ones (this can be tested with `pstree -ap` under linux) and eventually slowing down the system. Upon closing the python interpreter, all processes are terminated at once.

    import numpy as np
    import emcee

    import multiprocessing, logging
    mpl = multiprocessing.log_to_stderr()
    mpl.setLevel(logging.INFO)

    def lnprob(x, mu, icov):
        diff = x-mu
        return -np.dot(diff,np.dot(icov,diff))/2.0

    ndim = 50

    means = np.random.rand(ndim)

    cov = 0.5 - np.random.rand(ndim ** 2).reshape((ndim, ndim))
    cov = np.triu(cov)
    cov += cov.T - np.diag(cov.diagonal())
    cov = np.dot(cov,cov)

    icov = np.linalg.inv(cov)

    nwalkers = 250

    threads = 8

    for i in range(10):
        p0 = np.random.rand(ndim * nwalkers).reshape((nwalkers, ndim))
        sampler = emcee.EnsembleSampler(nwalkers, ndim, lnprob, args=[means, icov], threads=threads)
        pos, prob, state = sampler.run_mcmc(p0, 100)
        sampler.reset()
        # Do something with the samples produced...

This commit adresses this problem by adding a `weakref.finalize` within the `EnsembleSampler` class that calls the `InterruptiblePool.terminate`:

        self._pool_finalizer = weakref.finalize(self, InterruptiblePool.terminate, self.pool)

This way when the  `EnsembleSampler` is deleted (in the example above when a new instance is assigned to `sampler`), the associated pool is terminated.

dfm:
This is very cool! Thanks for the contribution. I'm not going to merge it, however, because v3.0 no longer has a `threads` argument. This simplifies all the parallel usage. Thanks again!

PR #233, opened 2017-10-05
use np.isinf instead of not np.isfinite

basnijholt:
I was working through the example and think that `np.isinf` is better than `not np.isfinite`.


dfm:
Why do you think that? `isfinite` also checks for NaNs which is the desired behavior. If there's some function that actually does `not np.isfinite`, I'd love to know about it.

basnijholt:
> `np.isfinite` also checks for NaNs which is the desired behavior

I didn't realize that! I'll close the PR.

dfm:
Thanks!

PR #234, opened 2017-10-06
Change xrange to range

basnijholt:
Where the code was Python 2 compatible I removed the try except for `xrange` because
`xrange(nwalker)` where `nwalkers=50` is hardly an optimization anyway.


dfm:
I'm actually in the process of re-writing all the docs and examples so this is something I'll definitely incorporate when I do that.

PR #241, opened 2017-11-07
MAINT: make sure iterations steps are always run

andyfaff:
As [outlined on the ML](https://groups.google.com/d/topic/emcee-users/LiBhiz-ZiCY/discussion)

### DIFF
--- a/emcee/ensemble.py
+++ b/emcee/ensemble.py
@@ -5,6 +5,7 @@ from __future__ import division, print_function
 __all__ = ["EnsembleSampler"]
 
 from collections import Iterable
+from itertools import cycle
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


andyfaff:
Thinking about this a bit more.
There must have been be a rationale for the new behaviour. However, the old/new behaviour (thin vs thin_by) has radically different behaviour w.r.t the total number of iterations carried out.

There is quite a lot of changed behaviours across the project (e.g. `EnsembleSampler.chain` and `EnsembleSampler.get_chain` have swapped axes) which make it hard for downstream packages to cope with old+new in a single codebase.
I'm wondering if it would be better to create an `emcee3` project instead. This would permit the widerange removal of all the deprecated behaviour. This would enable downstream packages to make a cleaner break to use the new behaviour. I for one would just use an emcee3, but I could envisage that others could continue to work with the existing package.

dfm:
This is basically the only real change and thanks to you it's totally
backwards compatible! The other features that were removed were full of
bugs and probably no one should be using them anyways. I do have an emcee3
project, but it probably won't ever be released. This was my compromise. I
love that you're asking me to support two separate libraries now!

The thin_by behavior is as expected - you always produce iterations
samples.

PR #247, opened 2017-11-22
BUG: compute_log_prob should accept coords=None

andyfaff:
The docstring for `compute_log_prob` indicates it should work with `coords=None`.  Reproducing code is:

```
import numpy as np
import emcee

def lnprob(x, ivar):
    return -0.5 * np.sum(ivar * x ** 2)

ndim, nwalkers = 10, 100
ivar = 1. / np.random.rand(ndim)
p0 = [np.random.rand(ndim) for i in range(nwalkers)]

sampler = emcee.EnsembleSampler(nwalkers, ndim, lnprob, args=[ivar])
sampler.run_mcmc(p0, 100)
sampler.compute_log_prob()
```

### DIFF
--- a/emcee/ensemble.py
+++ b/emcee/ensemble.py
@@ -351,7 +351,7 @@ class EnsembleSampler(object):
         """Calculate the vector of log-probability for the walkers
 
         Args:
-            pos: (Optional[ndarray[..., ndim]]) The position vector in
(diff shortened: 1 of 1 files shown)


dfm:
Good catch - thanks! This docstring was a holdover from the old version. I've fixed the docstring instead of merging this code.

PR #251, opened 2017-12-12
Clarify that vectorize disables pool

eteq:
Minor docstring item here, but it threw me for a loop while experimenting.  I *think* I'm interpreting the beahvior right, aren't I?


dfm:
Yes. Thanks!

PR #257, opened 2018-06-18
Fix wrong argument name in exception

MaxNoe:
The exception that is raised on `initial_state=None` on first run and the comment below still used `pos0`

### DIFF
--- a/emcee/ensemble.py
+++ b/emcee/ensemble.py
@@ -325,15 +325,17 @@ class EnsembleSampler(object):
         """
         if initial_state is None:
             if self._previous_state is None:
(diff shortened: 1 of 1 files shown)


dfm:
Awesome. Thanks!

MaxNoe:
So you have a timetable for the 3.0 release?

dfm:
I do not. I'm using it for all my projects already, but I'm having a hard time pulling the plug and calling it finished :-\ I'll try to make a specific plan soon!

MaxNoe:
> I'm using it for all my projects already

That's why i'm asking. A new PyPI release would be nice. You could release a 3.0.0rc1, so it's clear it's not yet finished, but people can install it using pypi.

PR #261, opened 2018-07-24
Make reset compatible with and demo multiple-runs-in-one-file hdf5 usage

eteq:
This PR does two related things:

1. Makes an adjustment to the ``reset`` in the hdf5 backend so that it doesn't overwrite the old file but rather just overwrites the corresponding group/datasets for the backend's name.
2. Demonstrates in the relevant doc section how, with this change, one can now safely store multiple ``emcee`` runs in the same hdf5 file.

One thing to note: in the `sampler2` block I didn't include the cleverness in choosing the end iteration on convergence and instead just used ``run_mcmc``. @dfm, if you feel that's not the way it should be, I can change that... But when I tried to use the pattern you followed for the first sampler it seemed like a distraction from the real point for that block, which is the ``name`` option of the backend and how it plays out.


eteq:
Oh and I see now that my editor automatically cleared out some of the whitespace that was previously in the blank lines.  If you want that left in, @dfm, I can go back and fix it (but won't bother if you don't care).

dfm:
This looks good. Can you update [the notebook](https://github.com/dfm/emcee/blob/master/docs/_static/notebooks/monitor.ipynb) instead of the rst? (The rst is generated automatically - that's where that whitespace is coming from).

Could you also add a test for this behavior? https://github.com/dfm/emcee/blob/master/tests/unit/test_backends.py

Thanks!

eteq:
Done as you suggested, @dfm - note I also made a small change in ``test_backends.py`` to use the ``tmpdir`` py.test fixture elsewhere in that file, since it was convenient to do that anyway for the test I added.  Can back that out if it's undesired, though.

 I also updated the rst from the notebook using your makefile.  I did *not* include the images it generated on the theory that that's a 1.3MB commit maybe you don't want in the repo.  But it means the html may be subtly inconsistent if my run didn't go exactly the same as the run that that image was previously generated in (they look virtually identical to me, but probably aren't at the pixel level)...

davidwhogg:
I'm loving @eteq's contributions to this project.

dfm:
Same. Thanks @eteq!

PR #262, opened 2018-07-24
Make progress bar optionally use other tqdm pbars

eteq:
This PR enhances the progress bar support to allow use of some of the other `tqdm` progress bars.

Most notably, this enables ``tqdm_notebook``, which I think is much better when working in the notebook environment.  It also allows ``tqdm_gui``, presumably, although I've not tested it.  And if `tqdm` adds more in the future, it should allow that as long as they follow the ``tqdm.tqdm_<something>`` convention for function names.


dfm:
Cool. Thanks!

PR #263, opened 2018-07-24
Fix quickstart

eteq:
This makes a one-line change that seems to be necessary for the quickstart tutorial to work in the current version. 

I'm guessing the old part is a hold-over from a previous version? I also updated the text to reflect the new situation of a `State` object instead of the previous tuple.


dfm:
Thanks Erik! I actually fixed this myself because I've switched to using notebooks for the docs so the updated needed to happen there. Thanks for pointing this out!

PR #268, opened 2018-08-07
State docstring

### LINKED ISSUE #264: Add a docstring for `State`/add to docs
I think this is pretty trivial (and might be able to get around to it myself), but want to make sure there's not some reason to avoid this (like `State` is still in flux).

Since `State` is now potentially important as the return (and possible input) of `sample`/`run_mcmc`, it probably makes sense to have it documented and get an ``autoclass`` somewhere in the docs, right?

eteq:
This closes #264 by adding a docstring for `State`, including it in the docs, and updating the various references to it to hyperlink to the right place.

@dfm you might want to confirm I got everything about `State` right in the new docstring - it's pretty straightforward but I admit I didn't check if anything there of relevance has changed between emcee 2 and 3...

### DIFF
--- a/docs/user/sampler.rst
+++ b/docs/user/sampler.rst
@@ -10,3 +10,10 @@ Standard usage of ``emcee`` involves instantiating an
 
 .. autoclass:: emcee.EnsembleSampler
    :inherited-members:
+
+Note that several of the :class:`EnsembleSampler` methods return or consume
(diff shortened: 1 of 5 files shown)


dfm:
Thanks! I don't understand what's going on with the appveyor failures, but it's not anything that you changed.

PR #270, opened 2018-08-09
Fix for numpy 1.15

eteq:
The appveyor failures in #268 piqued my interest even though they were not directly related to that PR.  And it reveals a failure that seems to be due to numpy: it turns out numpy 1.15 (which I can only assume just got picked up by appveyor recently) had a subtle change of behavior: for whatever reason `numpy.squeeze` no longer accepts integer arrays, and that approach is used in processing blobs in `emcee`.  But `squeeze`  seems to behave exactly the same if you just convert the axes argument to tuples...  So this PR just makes that trivial change.

Local testing on 1.15 seemed to do the job.  We shall see what appveyor says, though.

### DIFF
--- a/emcee/ensemble.py
+++ b/emcee/ensemble.py
@@ -406,7 +406,7 @@ class EnsembleSampler(object):
             if len(shape):
                 axes = np.arange(len(shape))[np.array(shape) == 1] + 1
                 if len(axes):
-                    blob = np.squeeze(blob, axes)
+                    blob = np.squeeze(blob, tuple(axes))
 
         # Check for log_prob returning NaN.
         if np.any(np.isnan(log_prob)):


manodeep:
multi-dim array indexing has changed in `1.15`  -- only tuples are allowed now. See the second item in [deprecations](https://docs.scipy.org/doc/numpy/release.html#deprecations)

eteq:
@manodeep - thanks for pointing that out! A bit annoying that they use the term "deprecation" even though it's not a deprecation but an immediate fail... Ah well.

Anyway, looks to have done the job.

manodeep:
@eteq Agreed. Deprecation should only yield a deprecation warning, and not trigger a test failure. Kind of the point for having deprecation.

andyfaff:
Sometimes one wants to have the `DeprecationWarning` create a CI test failure. Otherwise the deprecations can be missed, only to have code start failing in use because the changes weren't made early enough.

eteq:
That's may be true in the CI @andyfaff, but in this case it's an *actual* failure. That is, anyone who updates to numpy 1.15.0 will start seeing this failure if they are using blobs, I think.

Which begs a question for @dfm: are you planning to have a maintenance branch for emcee 2? If so I guess this should be backported there. If not it's a great advertisement for 3: "Now with numpy 1.15 compatibility!" :wink:

dfm:
Haha! Thanks for all this. I would like to have a maintenance branch so I'll have a go at backporting this.

dfm:
This actually won't be a problem for v2 because the treatment of blobs was much simpler back in the good ole days :-)

I've merged this in for v3. Thanks!

PR #283, opened 2018-11-28
MAINT: import from collections.abc

andyfaff:
In Py3.8 importing an ABC from `collections` will result in an Exception. In Py3.7 the behaviour is deprecated. This PR addresses the DeprecationWarning. This PR should be applied to the next `emcee` release, as that will be used extensively on Py3.8.

```
DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```


dfm:
Thanks!

PR #284, opened 2018-11-28
CI: include py37 in travisCI matrix

andyfaff:
This PR adds a py37 entry to the travisCI matrix. Testing py37 is important for finding issues such as that addressed by #283


dfm:
Thanks!

PR #304, opened 2019-08-09
DOC: randomize_split default True

andyfaff:
The docstring for `RedBlueMove` does not match the default kwd arg in `__init__`. This PR changes the docstring to reflect reality.


dfm:
Thanks!

Issue #306, opened 2019-09-08
Add check that `initial_state` allows full exploration of parameter space


Christopher-Bradshaw:
When running the ensemble sampler, I occasionally accidentally initialize all walkers to have the same initial position in some of the dimensions. This is entirely my fault! But, it would be easy (I think) to add a check that the initial state allows the sampler to explore all of parameter space, rather than have me find out my mistake after running my chains for a while.

If you think this would be a useful thing to have, I'm happy to make the change and submit a PR, but wanted to check it was something you'd be interested in/I haven't missed some reason why you might not want this.

Thanks!

davidwhogg:
The test is that the determinant of the variance tensor of the walker positions (minus the mean position) is nonzero. It's easy to implement.

davidwhogg:
ps. It will be violated when live_dangerously=True

Alexis-Prel:
This is what I am using for this check before I send `p0` to `emcee`. 

```python
degenerate = [p for p in range(dim) if np.unique(p0[:, p]) == 1]
assert not degenerate, f"Parameters {degenerate} can not evolve."
```

davidwhogg:
I don't think that check is quite right. Even if there are non-unique entries, you still might be fine.

Alexis-Prel:
Maybe I made a mistake! I'm not sure I understand your comment though.

My logic was that if `np.unique(p0[:, p])` has only one element `u`, then all stretch moves will return the same position because the proposal is `q = c[rint] - (c[rint] - s) * zz[:, None]`  and then `q[p] = u - (u - u) * zz[p] = u`.

davidwhogg:
Okay got it! I did mis-read the code. But still, I think that the right check is something like
```
  dp0 = p0 - np.mean(p0, axis=0)[None, :]
  var = (1. / nwalkers) * np.sum(dp[:, None, :] * dp[:, :, None], axis=0)
  sgn, logdet = np.linalg.slogdet(var)
  assert logdet > 0.
```
or maybe something more robust.

Christopher-Bradshaw:
@Alexis-Prel that is actually roughly what I had in mind before @davidwhogg made his suggestion. But after thinking about it, I don't think that is enough. Consider in a 2d space `p0 = [(1, 1), (2, 2), (3, 3), ... (n, n)]`.

This passes your test, but using the stretch move, those walkers can never explore anything other than the `x = y` line. Let me know if that makes sense or if I have missed something!

[...]

PR #307, opened 2019-09-11
Print a warning if initial state is in a hyperplane

Christopher-Bradshaw:
Mentioned this in https://github.com/dfm/emcee/issues/306. Use case is that I occasionally accidentally initialize all my walkers at the same location in one or more axes. Rather than find out about that after running my chains, this will warn about it when they start.

Let me know if you want any changes made, or feel free to modify as you see fit.


dfm:
Thanks for this! Do you think you could also add an option to disable this check? And a test to make sure that it *doesn't* raise the warning when it shouldn't?

dfm:
PS the failed tests are unrelated to this pull request so, if you rebase on master, it should all pass.
Thanks!

Christopher-Bradshaw:
Definitely a good idea to check that it doesn't warn when it shouldn't! This should be good to go now, but I'll check back in a bit to make sure tests pass.

Let me know if there is anything else to do

PR #314, opened 2019-10-17
Scale-invariant Test for Linear Dependence

farr:
At the beginning of `run_mcmc`, the sampler object tests for linear dependence of the initial state.  This pull request computes the condition number of the covariance matrix of the initial state, and compares against `1e8`.  The previous code computed the determinant of the covariance and compared to zero; that method is sensitive to the *scale* of the parameters (i.e. under re-scaling of parameters `x` to `x_prime = A*x`, the determinant scales as `det(cov(x))` to `det(cov(x_prime)) == A**2*det(cov(x))`.  In a parameter space with small parameters, `det(cov(x))` can be close to zero even when the parameters are quite linearly independent, leading to spurious warnings.

### DIFF
--- a/emcee/ensemble.py
+++ b/emcee/ensemble.py
@@ -244,14 +244,9 @@ class EnsembleSampler(object):
         state = State(initial_state, copy=True)
         if np.shape(state.coords) != (self.nwalkers, self.ndim):
             raise ValueError("incompatible input dimensions")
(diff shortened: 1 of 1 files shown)


dfm:
Thanks!

farr:
My pleasure.

PR #315, opened 2019-10-17
Quick start ipynb: Text inconsistent with code cell.

hamogu:
It looks to me as if the variable `run` that is referenced in the test, should really be the variable `state`. Maybe the name has been changed in the past without changing the line of text? (Alternatively, I might just not understand what's going on there and where the variable `run` is supposed to come from.)


dfm:
You're right - Thanks!

Issue #312, opened 2019-10-02
emcee 3.0 breaks extended-precision coordinates


aarchiba:
We are working with variables of type np.longdouble (hardware 80-bit floating-point, stored in np.float96 or np.float128 depending on platform), and the switch from emcee 2.2 to emcee 3.0 breaks using these as walker coordinates. Specifically, np.linalg is not capable of doing linear algebra on these extended-precision values, but scipy.linalg is, and sample() calls np.linalg.det on a covariance matrix constructed from these coordinates.

One option would be to try to import linalg from scipy, and fall back to numpy if this is not available; this would have a higher chance of using faster system BLAS packages too, and it's reasonable to fail to work if no high-precision implementation is available (we require scipy anyway). 

If the determinant is an approximate value, for example it is applied to a covariance matrix for a small sample, then it might be harmless and simple to just apply .astype(float) to the covariance matrix. Of course determinants can be numerically precarious, but the determinant call is checking for approximate linear dependence, so it's probably just fine to approximate. 

We are using this in PINT ( https://github.com/nanograv/PINT ) which works with time spans of decades that can be measured to the nanosecond, so we really do need the extended precision, though often by the time we are down to the quantities of real interest the need for precision is less.

dfm:
Thanks for this update! I think that a good temporary workaround would be to pass `skip_initial_state_check = True` to sample, but I'll try out your proposed options for a better long term solution.

aarchiba:
If I were to write up a few test cases to confirm that emcee worked okay with extended precision, would that be a good addition to the test suite? Nothing too elaborate.

dfm:
That would be wonderful if you get a chance!

aarchiba:
Turns out the assumption that walker coordinates are floats is baked into the chain storage. Do you want to also allow complex coordinates? I mean I think the affine thing should work fine there, so it's just a matter of allowing the dtype?

aarchiba:
(Incidentally although https://github.com/aarchiba/triplesystem used emcee, I was using downhill optimizers as well and so I handed them all normalized coordinates that were of about zero mean and order unity standard deviation. It sure didn't simplify things any, though.)

dfm:
Interesting. Let me think a bit more about complex coordinates, but I think it definitely makes sense to support different dtypes for the coordinates in general. I'll look into how much that'll take to do.

aarchiba:
[...]

PR #319, opened 2019-10-21
Support  longdouble data types

aarchiba:
The goal is to have emcee work with coordinates that are numpy long doubles (80-bit precision on most Intel, 128-bit on arm64, 64-bit on MSVC). It's not essential that affine combinations permit long doubles as coefficients; the concern is when the clump of walkers is in a small region around some large-ish long double value. This comes up when fitting times in pulsar data.


aarchiba:
Current state is, as far as I can tell, that somewhere the HDF5 backend eats the extra precision. I was planning to write a series of backend tests that walk through opening, creating a new dataset, adding things, closing, and reopening, and at each step checking data types, both returned and in the HDF5 file.

aarchiba:
Addresses #312

dfm:
So it looks like most of the errors are actually coming from the linear dependence check. Note that `cond` doesn't exist in `scipy.linalg` so it'll take a bit more tweaking to get that to work. Perhaps we should just skip the check if the dtype isn't `float32` or `float64`?

Also: I recently rearranged the repo a bit to put all the code in a `src` directory. Could you merge the current master and move the new tests to `src/emcee/tests/unit` instead of `emcee/tests/integration`?

aarchiba:
I did the merge but forgot to move the new test file. I'll do that.

aarchiba:
The errors coming from the check are easy to fix or ignore. I'm afraid the more serious error may have gotten lost in the merging, I'll try to surface it again, but:

with a default backend, long doubles actually  work as long as you pass in dtype=np.longdouble.

With the HDF5 backend, long doubles appear to work but the extra precision is lost somewhere, even though no exception is raised and I create all datasets with a dtype of np.longdouble.

aarchiba:
Well. There's an unfortunate partial explanation: HDF5 has no portable IEEE 80-bit float type. No portable floating-point type with more than 64 bits at all. The standard does permit native long doubles, but I'm not sure of the status of h5py's support. I'll look into this.

aarchiba:
Am I mistaken or did you blacken the source code? That is, can I just run black willy-nilly and stop worrying about formatting?

dfm:
Yes I did! The only custom configuration is [the line length defined here](https://github.com/dfm/emcee/blob/master/pyproject.toml#L5) but black should find that automatically!

aarchiba:
Nothing so grand I just forgot to pass dtype from the temporary HDF5 thing to the actual one.

Is it worth warning users that the generated HDF5 files are non-portable if they contain long doubles? In particular arm64 has quadruple precision long doubles. (Travis sort of supports this architecture now, if you want to add tests; on Windows long double is the same as double, so funny things can happen there too, but I think my code is fine if pointless there.)

I don't support complex values, and I use 64-bit floats for the condition number - but I added a normalization step which should help everyone.

aarchiba:
> Yes I did! The only custom configuration is [the line length defined here](https://github.com/dfm/emcee/blob/master/pyproject.toml#L5) but black should find that automatically!

Very nice! I just did that to PINT, which has the usual problem of being written by expert scientists who are not expert programmers. I added a pre-commit hook and a Travis job to check continued blackness: https://github.com/nanograv/PINT/commit/ef448d5271e7ce4a2de912dd2a76a629e7a569cb and also later a hack so that modern git blame will ignore the blackening commit: https://github.com/nanograv/PINT/commit/12ea9cca3d26be549807259cd325ef31d678d3f2 plus documentation on how to set up your environment to use these: https://nanograv-pint.readthedocs.io/en/latest/development-setup.html#command-line-tools-and-automation

dfm:
Very nice! If you have ipynbs, I made a project to enable black for them as a check or git hook: https://github.com/dfm/black_nbconvert

I'm using that for [my current main project](https://github.com/dfm/exoplanet/blob/master/.github/workflows/codestyle.yml), but I don't have any developer docs yet...

aarchiba:
We are using [jupytext](https://github.com/mwouts/jupytext) to manage our notebooks in git - only markdown versions with no output are stored, so diffing and merging is easy. Then [nbsphinx](https://nbsphinx.readthedocs.io/en/0.4.3/) runs and renders them into our online docs, and `jupytext --pipe black --pipe-format py:percent` handles blackening.

aarchiba:
It might have gotten lost in the conversation but: it appears that long doubles now work.

aarchiba:
Would you be interested in a Travis job on arm64, where long double is binary128 (quad precision)? I have astropy building and passing the test suite but arm64 is still a bit experimental and slow. (I don't currently have access to an arm64 machine for testing.)

dfm:
Thanks for this!

PR #320, opened 2019-10-27
Add Travis arm64 jobs

aarchiba:
On arm64 long double is quadruple precision; in combination with #319 this should avoid surprises. Ideally we would complement this with a test on Windows, where long doubles are plain binary64, the same as doubles.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,11 @@ env:
   - NUMPY_VERSION=1.17
   - NUMPY_VERSION=1.16
   - NUMPY_VERSION=1.15
+jobs:
+  include:
+    - python: "3.7"
+      arch: arm64
+      env: NUMPY_VERSION=1.17
 
 install:
   - travis_retry pip install --upgrade pip


dfm:
Thanks! This is currently failing on the numpy build because of missing blas/lapack routines. I'll take a look at it later today if I get a chance!

aarchiba:
Don't worry too much about this - the Travis arm64 setup is still very experimental, and missing dependencies are a known headache. I can probably fix it by apt-getting the appropriate blas/lapack Ubuntu packages, but it will still be building numpy and scipy from source, which will take approximately forever. While in principle the pip cache might solve the problem after the first build, it still doesn't seem to work right and might even be trying to share binary wheels between arm64 and amd64. I might have better luck if I can rely upon system numpy and scipy packages.

dfm:
OK - thanks! Let me know if you make any progress or if there's anything I can help with.

aarchiba:
I have an arm64/aarch64 machine now while and the absence of compiled wheels is very annoying emcee's tox passes fine. I might write a few more stressful extended precision tests but it looks like we're safe using quadruple precision when it's available.

Issue #321, opened 2019-10-29
emcee won't run from source


tlatorre-uchicago:
**General information:**

- emcee version: master
- platform: Linux
- installation method (pip/conda/source/other?): source

**Problem description:**

When trying to install emcee from source it seems that the emcee_version.py file never gets created, and so I'm getting an ImportError.

### Expected behavior:

```python
>>> import emcee
```

### Actual behavior:

```python
>>> import emcee
ImportError: No module named emcee_version
```

### What have you tried so far?:

I grepped through the code base looking for an emcee_version.py file and noticed that it's in the .gitignore, so I guess it's a file which isn't stored in the git repository.

I bisected the problem to 13595f7c04f6e0711cc34567e398576e375043bc which is where the emcee_version.py file was introduced.

### Minimal example:

```bash
$ git clone https://github.com/dfm/emcee
$ cd emcee
$ python setup.py install
$ cd ~
```
and then try to import it:

```python
>>> import emcee
```

dfm:
This is probably because `setuptools_scm` must be installed before running `setup.py`. What happens if you do:

```bash
pip install .
```

If that doesn't work,

```bash
pip install -U setuptools setuptools_scm
python setup.py install
```

should do the trick. Let me know and we'll update the docs!

tlatorre-uchicago:
Just trying pip install fails with:

```bash
(emcee_env) [tlatorre@localhost emcee]$ pip install .
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
Processing /home/tlatorre/emcee
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... error
    ERROR: Command errored out with exit status 1:
     command: /home/tlatorre/emcee_env/bin/python2 /home/tlatorre/emcee_env/lib/python2.7/site-packages/pip/_vendor/pep517/_in_process.py prepare_metadata_for_build_wheel /tmp/tmp3XO2W1
         cwd: /tmp/pip-req-build-ObXBAv
    Complete output (2 lines):
    running dist_info
    error: 'egg_base' must be a directory name (got `/tmp/pip-modern-metadata-GV2AR8`)
    ----------------------------------------
[...]

PR #322, opened 2019-10-29
update setup.py to fix building from source

tlatorre-uchicago:
This commit updates setup.py to add:

    setup(
        ...,
        use_scm_version=...,
        setup_requires=['setuptools_scm'],
        ...,
    )

as recommended by the setuptools-scm project at
https://pypi.org/project/setuptools-scm/.

This commit fixes builds from source which were previously broken due to an ImportError:

    $ git clone https://github.com/dfm/emcee
    $ cd emcee
    $ python setup.py install
    $ cd ~
    $ python
    $ import emcee
    ImportError: No module named emcee_version

This should fix issue #321.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,6 +53,7 @@ if __name__ == "__main__":
             ),
             "write_to_template": '__version__ = "{version}"\n',
         },
+        setup_requires=['setuptools_scm'],
         author=find_meta("author"),
         author_email=find_meta("email"),
         maintainer=find_meta("author"),


dfm:
That means that your setuptools must be out of date because `setup_requires` is already defined in `pyproject.toml` and I don't want to define it multiple places!

tlatorre-uchicago:
Huh... ok. Yeah here is the version of setuptools I'm using (Fedora 27):

```python
>>> import setuptools
>>> setuptools.__version__
'37.0.0'
```

tlatorre-uchicago:
Actually wait, nm. I was installing in a virtualenv with what appears to be the latest version of setuptools:

```python
>>> import setuptools
>>> setuptools.__version__
'41.6.0'
```

dfm:
Strange. I'll look into it. Just to confirm: you're on py2.7 right?

tlatorre-uchicago:
Yep

```
[tlatorre@localhost ~]$ python
Python 2.7.15 (default, Oct 15 2018, 18:36:25) 
[GCC 7.3.1 20180712 (Red Hat 7.3.1-6)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
```

dfm:
I bet that's the issue and I'm not going to sweat too much about Python 2 support since numpy is no longer supported so can you double check that running:

```bash
pip install -U setuptools_scm
python setup.py install
```

works for you?

I've already added a test to travis that makes sure that the current configuration works for py3.6 and 3.7 on Travis.

dfm:
Interestingly it seems to work fine with 2.7 on Travis too. I'm not too sure what the issue on your system is!

tlatorre-uchicago:
Yeah it looks like upgrading setuptools_scm did fix it. I have no idea. Python packaging is a bit of a mess. It's very hard for even me to reproduce the issue (to reproduce the original issue I have to wipe out the whole site-packages in the virtual environment and then delete the following directories in the emcee git repo: dist/ build/ src/emcee.egg_info /src/emcee_version.py).

So, maybe it was just setuptools_scm? I'm not even sure how to find out the version of that.

dfm:
Yeah - that's annoying! I've added a few words to the docs to try to make this simpler for the next time: https://emcee.readthedocs.io/en/latest/user/install/

Thanks for looking into this!

Issue #313, opened 2019-10-05
How can emcee be run deterministically?


aarchiba:
I want to be able to run emcee multiple times and get exactly the same answer. How can this be accomplished? It does not seem to be sufficient to use random.seed(0), or numpy.random.seed(0), or to capture numpy.random.mtrand.RandomState(0) and assign it to the sampler, or in fact all three.

This is relevant, among others, for making sure that our example notebook doesn't change when rerun, so that it can be checked in to version control without endless merge conflicts.

A quick test in raw emcee; this passes, mystifyingly, sometimes but not always:
```python
def test_raw_emcee():
    r = []
    for i in range(2):
        random.seed(0)
        numpy.random.seed(0)
        s = numpy.random.mtrand.RandomState(0)

        def log_prob(x, ivar):
            return -0.5 * np.sum(ivar * x ** 2)

        ndim, nwalkers = 5, 100
        ivar = 1. / np.random.rand(ndim)
        # r.append(ivar[0])
        p0 = np.random.randn(nwalkers, ndim)
        #r.append(p0[0, 0])

        sampler = emcee.EnsembleSampler(nwalkers, ndim, log_prob, args=[ivar])
        sampler.random_state = s
        sampler.run_mcmc(p0, 100)

        samples = sampler.chain.reshape((-1, ndim))
        r.append(samples[0, 0])
    assert r[0] == r[1]
```

dfm:
I would expect this to work and it does on my machine (although I only ran it ~5000 times). How frequently do you see the failure and what version of numpy do you have?

aarchiba:
> I would expect this to work and it does on my machine (although I only ran it ~5000 times). How frequently do you see the failure and what version of numpy do you have?

I see the failures a lot more often than that; I think the erratic behaviour actually just my fault, that the initial guesses are the same and so I see a failure if and only if walker 0 made a jump. 

My numpy is 1.17.2. Obviously I need to do some more investigation. I'll get back to you with narrower parameters.

It would be nice if there were an explicit "seed" or "state" argument so one could be sure no unauthorized randomness was produced inside the constructor before I have access to set the random state.

aarchiba:
Okay so I made this up into a file and in a virtualenv with emcee 3.0.0 it works fine; if I downgrade to emcee 2.2.1, changing nothing else, I get failure.

Alexis-Prel:
[...]

Issue #300, opened 2019-05-31
Potential bug in run_mcmc of the 2.2.x branch


AlexisPrel:
**General information:**

- emcee version: 2.2.x

**Problem description:**
Reading the [code for the Sampler class](https://github.com/dfm/emcee/blob/v2.2.x/emcee/sampler.py) of version 2.2.x, I think I found a mistake at line **167** in `sampler.py`: 

Shouldn't `rstate0` be replaced with `lnprob0` ?

dfm:
Yes! You're absolutely right.

PR #324, opened 2019-11-02
V2.2.x

Alexis-Prel:
This addresses issues #300 and #313, including the suppression of direct references to np.random. 
It also adds a new test case  `test_deterministic` to demonstrate that re-running `run_mcmc` with the same random state produces the exact same output.

### DIFF
--- a/emcee/ptsampler.py
+++ b/emcee/ptsampler.py
@@ -7,7 +7,6 @@ from __future__ import (division, print_function, absolute_import,
 __all__ = ["PTSampler"]
 
 import numpy as np
-import numpy.random as nr
 import multiprocessing as multi
 
 from . import autocorr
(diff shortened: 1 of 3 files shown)


emcee/ptsampler.py
```diff
@@ -165,6 +164,10 @@ def __init__(self, ntemps, nwalkers, dim, logl, logp, threads=1,
         self.nwalkers = nwalkers
         self.dim = dim
 
+        # This is a random number generator that we can easily set the state
+        # of without affecting the numpy-wide generator
+        self._random = np.random.mtrand.RandomState()
+
```
Alexis-Prel: This duplicates code from `sampler.Sampler.__init__`
Not sure if that's a maintenance issue ...

Alexis-Prel:
I get this test failure 

```
test_nan_lnprob ...
Failed with:
    TypeError: ufunc 'isinf' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

But if I am not mistaken, this test also fails with the current version of v2.2.x

dfm:
Thanks! Where are you seeing that test failure? It looks like the failure on Travis is unrelated and the actual unit tests run just fine.

dfm:
Wait - I see it now. The Python 2.7 test is failing on Travis too. Looks like 3.4 and 3.5 are just fine. I'm happy to merge this and we can look into that issue elsewhere. Thanks!

PR #325, opened 2019-11-03
Adding packaging badge under installation

migueldvb:
Including the repology packaging badge in the installation list. This gives a list of emcee package versions for project distribution (Linux, BSD) in all repositories. It can be useful information for people using those systems.

### DIFF
--- a/docs/user/install.rst
+++ b/docs/user/install.rst
@@ -13,7 +13,7 @@ All you'll need `numpy <https://numpy.org/>`_.
 Package managers
 ----------------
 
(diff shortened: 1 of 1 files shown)


dfm:
Thanks! I'd rather add a comment that prefers installation via `pip` or `conda` over these and over the source installation. Perhaps we should also make this a separate section?

migueldvb:
Sure! I included those changes in the commit.

dfm:
Perfect - thanks!!

PR #326, opened 2019-11-04
Fix and test linear dependence check

### LINKED ISSUE #323: [3.0.1] Test failures in test_sampler.py
**General information:**

- emcee version: 3.0.1
- platform: Debian unstable/testing
- installation method (pip/conda/source/other?): pip source

**Problem description:**

During packaging the new version for Debian, I get the following test errors:
```
_____________________________ test_errors[Backend] _____________________________

backend = <class 'emcee.backends.backend.Backend'>, nwalkers = 32, ndim = 3
nsteps = 5, seed = 1234

    @pytest.mark.parametrize("backend", all_backends)
    def test_errors(backend, nwalkers=32, ndim=3, nsteps=5, seed=1234):
[…]
    
            # Ensure that a warning is logged if the inital coords don't allow
            # the chain to explore all of parameter space, and that one is not
            # if we explicitly disable it, or the initial coords can.
            with pytest.warns(RuntimeWarning) as recorded_warnings:
>               sampler.run_mcmc(np.ones((nwalkers, ndim)), nsteps)

emcee/tests/unit/test_sampler.py:120: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
emcee/ensemble.py:381: in run_mcmc
    for results in self.sample(initial_state, iterations=nsteps, **kwargs):
emcee/ensemble.py:247: in sample
    np.atleast_2d(np.cov(state.coords, rowvar=False))
emcee/ensemble.py:558: in _scaled_cond
    return np.linalg.cond(c.astype(float))
/usr/lib/python3/dist-packages/numpy/linalg/linalg.py:1716: in cond
    s = svd(x, compute_uv=False)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

a = array([[nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan]])
full_matrices = True, compute_uv = False

    @array_function_dispatch(_svd_dispatcher)
    def svd(a, full_matrices=True, compute_uv=True):
      """[…]"""
        a, wrap = _makearray(a)
        _assertRankAtLeast2(a)
        t, result_t = _commonType(a)
    
        extobj = get_linalg_error_extobj(_raise_linalgerror_svd_nonconvergence)
    
        m, n = a.shape[-2:]
        if compute_uv
            […]
        else:
            if m < n:
                gufunc = _umath_linalg.svd_m
            else:
                gufunc = _umath_linalg.svd_n
    
    
            signature = 'D->d' if isComplexType(t) else 'd->d'
>           s = gufunc(a, signature=signature, extobj=extobj)
E           ValueError: On entry to DLASCL parameter number 5 had an illegal value
/usr/lib/python3/dist-packages/numpy/linalg/linalg.py:1624: ValueError
```
and the same for `TempHDFBackend`.
Numpy version is 1.16.5, Python 3.7.5, x86_64.
This is a regression; 3.0.0 does not show this failure in the same build environment.

aarchiba:
Fixes #323

### DIFF
--- a/src/emcee/__init__.py
+++ b/src/emcee/__init__.py
@@ -22,11 +22,12 @@ __description__ = "The Python ensemble sampling toolkit for MCMC"
 from .emcee_version import __version__  # isort:skip
 
 from . import autocorr, backends, moves
-from .ensemble import EnsembleSampler
+from .ensemble import EnsembleSampler, walkers_independent
 from .state import State
 
 __all__ = [
     "EnsembleSampler",
+    "walkers_independent",
     "State",
     "moves",
     "autocorr",
--- a/src/emcee/ensemble.py
+++ b/src/emcee/ensemble.py
@@ -78,13 +78,19 @@ class EnsembleSampler(object):
     ):
         # Warn about deprecated arguments
         if a is not None:
-            deprecation_warning("The 'a' argument is deprecated, use 'moves' instead")
+            deprecation_warning(
+                "The 'a' argument is deprecated, use 'moves' instead"
+            )
         if threads is not None:
             deprecation_warning("The 'threads' argument is deprecated")
         if runtime_sortingfn is not None:
-            deprecation_warning("The 'runtime_sortingfn' argument is deprecated")
+            deprecation_warning(
+                "The 'runtime_sortingfn' argument is deprecated"
+            )
         if live_dangerously is not None:
-            deprecation_warning("The 'live_dangerously' argument is deprecated")
+            deprecation_warning(
+                "The 'live_dangerously' argument is deprecated"
+            )
 
         # Parse the move schedule
         if moves is None:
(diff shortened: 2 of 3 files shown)


aarchiba:
Ack, sorry for the reformatting in this commit, I don't know why black missed your line length setting. I'll fix it.

aarchiba:
There is some question about what the threshold for condition number should be; 1e8 seems a little conservative - for linear operators, the rule of thumb is that you should expect to lose eight decimal places above and beyond what round-off would do. For a clump of walkers... I'm not so sure. Once you've rescaled the clump to have mean zero and standard deviation one in each dimension, it is roughly the aspect ratio of the ellipsoid they form (if they're ellipsoidally distributed). So I guess it permits correlations of roughly 1-1e-16? That seems like a reasonable limit, I guess?

dfm:
This seems great for now - thanks!! My thought is to remove this from the sampler and expose it as a separate check that a user can use if they want before sampling. That way, people can protect themselves, but we'll have fewer issues with people running into this unnecessarily.

aarchiba:
Well, you'll notice the patch actually does export the test as an externally visible function, though I didn't add it to the documentation. And I think the way it's implemented now is a pretty solid test - I'll be surprised if we see false positives.

dfm:
I did see that 😄! Awesome - in that case let's leave it for now and see what happens.

dfm:
I guess one other thought is that not all "moves" require this condition to be satisfied. In the long run, we should probably take that into account as well...

PR #332, opened 2019-11-29
Add seed argument to EnsembleSampler

OriolAbril:
I saw this come up on the mailing list and had some time for this. I hope it helps.

This PR adds `seed` as a new argument to `EnsembleSampler` with default value `None`. Also, with `seed=None` the random state is not reseeded but instead it is left as is so that calling `np.random.seed()` before creating the sampler still works the same way. It also adds a test to check that the `seed` argument "overwrites" `np.random.seed`.

### DIFF
--- a/src/emcee/ensemble.py
+++ b/src/emcee/ensemble.py
@@ -58,6 +58,7 @@ class EnsembleSampler(object):
             to accept a list of position vectors instead of just one. Note
             that ``pool`` will be ignored if this is ``True``.
             (default: ``False``)
+        seed (Optional[int]): Seed for the random number generator.
 
     """
 
@@ -73,6 +74,7 @@ class EnsembleSampler(object):
         backend=None,
         vectorize=False,
         blobs_dtype=None,
+        seed=None,
         # Deprecated...
         a=None,
         postargs=None,
@@ -149,6 +151,8 @@ class EnsembleSampler(object):
         # of without affecting the numpy-wide generator
         self._random = np.random.mtrand.RandomState()
         self._random.set_state(state)
+        if seed is not None:
+            self._random.seed(seed)
 
         # Do a little bit of _magic_ to make the likelihood call with
         # ``args`` and ``kwargs`` pickleable.
--- a/src/emcee/tests/unit/test_sampler.py
+++ b/src/emcee/tests/unit/test_sampler.py
@@ -137,7 +137,8 @@ def run_sampler(
 ):
     np.random.seed(seed)
     coords = np.random.randn(nwalkers, ndim)
-    sampler = EnsembleSampler(nwalkers, ndim, normal_log_prob, backend=backend)
+    np.random.seed(None)
+    sampler = EnsembleSampler(nwalkers, ndim, normal_log_prob, backend=backend, seed=seed)
     sampler.run_mcmc(
         coords,
         nsteps,
@@ -319,3 +320,25 @@ def test_walkers_independent_randn_offset_longdouble(nwalkers, ndim, offset):
         np.random.randn(nwalkers, ndim)
         + np.ones((nwalkers, ndim), dtype=np.longdouble) * offset
     )
+
+
+def test_sampler_seed():
+    nwalkers=32
+    ndim=3
+    nsteps=25
+    np.random.seed(456)
+    coords = np.random.randn(nwalkers, ndim)
+    sampler1 = EnsembleSampler(nwalkers, ndim, normal_log_prob, seed=1234)
+    sampler2 = EnsembleSampler(nwalkers, ndim, normal_log_prob, seed=2234)
+    sampler3 = EnsembleSampler(nwalkers, ndim, normal_log_prob, seed=1234)
+    for sampler in (sampler1, sampler2, sampler3):
+        sampler.run_mcmc(
+            coords,
+            nsteps,
+        )
+    for k in ["get_chain", "get_log_prob"]:
+        a = getattr(sampler1, k)()
+        b = getattr(sampler2, k)()
+        c = getattr(sampler3, k)()
+        assert not np.allclose(a, b), "inconsistent {0}".format(k)
+        assert np.allclose(a, c), "inconsistent {0}".format(k)
--- a/src/emcee/tests/unit/test_sampler.py
+++ b/src/emcee/tests/unit/test_sampler.py
@@ -138,7 +138,9 @@ def run_sampler(
     np.random.seed(seed)
     coords = np.random.randn(nwalkers, ndim)
     np.random.seed(None)
-    sampler = EnsembleSampler(nwalkers, ndim, normal_log_prob, backend=backend, seed=seed)
+    sampler = EnsembleSampler(
+        nwalkers, ndim, normal_log_prob, backend=backend, seed=seed
+    )
     sampler.run_mcmc(
         coords,
         nsteps,
@@ -323,19 +325,16 @@ def test_walkers_independent_randn_offset_longdouble(nwalkers, ndim, offset):
 
 
 def test_sampler_seed():
-    nwalkers=32
-    ndim=3
-    nsteps=25
+    nwalkers = 32
+    ndim = 3
+    nsteps = 25
     np.random.seed(456)
     coords = np.random.randn(nwalkers, ndim)
     sampler1 = EnsembleSampler(nwalkers, ndim, normal_log_prob, seed=1234)
     sampler2 = EnsembleSampler(nwalkers, ndim, normal_log_prob, seed=2234)
     sampler3 = EnsembleSampler(nwalkers, ndim, normal_log_prob, seed=1234)
     for sampler in (sampler1, sampler2, sampler3):
-        sampler.run_mcmc(
-            coords,
-            nsteps,
-        )
+        sampler.run_mcmc(coords, nsteps)
     for k in ["get_chain", "get_log_prob"]:
         a = getattr(sampler1, k)()
         b = getattr(sampler2, k)()


andyfaff:
Could I suggest following the scipy approach:

```
    seed : int or `np.random.RandomState`, optional
        If `seed` is not specified the `np.RandomState` singleton is used.
        If `seed` is an int, a new `np.random.RandomState` instance is used,
        seeded with seed.
        If `seed` is already a `np.random.RandomState instance`, then that
        `np.random.RandomState` instance is used.
        Specify `seed` for repeatable minimizations.
```

The code typically uses [def check_random_state(seed)](https://github.com/scipy/scipy/blob/master/scipy/_lib/_util.py#L174):

```
# copy-pasted from scikit-learn utils/validation.py
def check_random_state(seed):
    """Turn seed into a np.random.RandomState instance
    If seed is None (or np.random), return the RandomState singleton used
    by np.random.
    If seed is an int, return a new RandomState instance seeded with seed.
    If seed is already a RandomState instance, return it.
    If seed is a new-style np.random.Generator, return it.
    Otherwise, raise ValueError.
    """
    if seed is None or seed is np.random:
        return np.random.mtrand._rand
    if isinstance(seed, (numbers.Integral, np.integer)):
        return np.random.RandomState(seed)
    if isinstance(seed, np.random.RandomState):
        return seed
    try:
        # Generator is only available in numpy >= 1.17
        if isinstance(seed, np.random.Generator):
            return seed
    except AttributeError:
        pass
    raise ValueError('%r cannot be used to seed a numpy.random.RandomState'
                     ' instance' % seed)
```

dfm:
@OriolAbril and @andyfaff: thanks for this! What would you think about breaking this out into its own function (perhaps called `check_random_state`) instead of having it embedded in the init?

OriolAbril:
I will get to it whenever I have time.

dfm:
@OriolAbril: Thanks - no rush!

OriolAbril:
Eventually got to it. How does this look?

dfm:
Thanks for your patience! This looks great. Perhaps we should add a reference to the sklearn implementation in the docstring?

dfm:
Thanks again for your patience! One more request:

Can you update the test to test the other two branches in the new code. I think that something like the following would work:

```python
sampler1 = EnsembleSampler(nwalkers, ndim, normal_log_prob, seed=1234)
sampler2 = EnsembleSampler(nwalkers, ndim, normal_log_prob, seed=sampler1._random)
```

andyfaff:
Sorry for just chipping in now. You'll see from `_check_random_state` that you can also supply a newer `np.random.Generator` instance as well as the legacy `np.random.RandomState`. The new `Generator` instances are higher performance, https://docs.scipy.org/doc/numpy/reference/random/index.html.

This means that the docstring for EnsembleSampler should (could) be:
```
seed (Optional[Union[int, np.random.RandomState, np.random.Generator]])
```

We've converted scipy over to enable the use of `Generator`. There the docstrings are:

```
    seed : {int, `~np.random.RandomState`, `~np.random.Generator`}, optional
        If `seed` is not specified the `~np.random.RandomState` singleton is
        used.
        If `seed` is an int, a new ``RandomState`` instance is used,
        seeded with seed.
        If `seed` is already a ``RandomState`` or a ``Generator`` instance,
        then that object is used.
        Specify `seed` for repeatable minimizations
```

Note that the methods for generating random numbers [have now changed slightly](https://docs.scipy.org/doc/numpy/reference/random/index.html#quick-start) and this will require some searching through the code base. e.g there is no `RandomState.integers` method anymore, so something like this will need to be used.

```
try:
    rg_integers = rg.integers
except AttributeError:
    rg_integers = rg.randint
a = rg_integers(1000)
```

However, the code as it's currently written will still work well, there are only going to be issues if people try to supply their own `Generator` instances.

OriolAbril:
I'll take a look at that and add an extra test with the generator

dfm:
Thanks both!

dfm:
@OriolAbril: This looks great to me now! Thank you so much for putting this together. @andyfaff let us know if you have any other suggestions and, if not, I'll merge.

andyfaff:
My only suggestion would be that the codebase be gone through and modified to work with np.random.Generator as well as np.random.RandomState

OriolAbril:
I tried to go through the codebase and update code not compatible with np.random.Generator and added a unit test.

I am not sure if this will be enough nor how should correct behaviour with Generator be tested, some changes are in moves code and are not checked in the one unit test added. Any suggestions will be most welcome.

src/emcee/moves/stretch.py
```diff
@@ -24,10 +24,18 @@ def __init__(self, a=2.0, **kwargs):
...
     def get_proposal(self, s, c, random):
+        try:
+            rg_integers = random.integers
+        except AttributeError:
+            rg_integers = random.randint
         c = np.concatenate(c, axis=0)
         Ns, Nc = len(s), len(c)
         ndim = s.shape[1]
-        zz = ((self.a - 1.0) * random.rand(Ns) + 1) ** 2.0 / self.a
+        try:
```
andyfaff: Instead of this try/except block you could use `random.uniform`.

```
zz = ((self.a - 1.0) * random.uniform(size=Ns) + 1) ** 2.0 / self.a
```
`uniform` also accepts the `low` and `high` keywords, so you might be able to simplify the term in the parenthesis.

andyfaff: I think it simplifies to (needs checking):
```
zz = (random.uniform(low=1, high=self.a, size=Ns)) ** 2.0 / self.a
```

src/emcee/moves/red_blue.py
```diff
@@ -93,11 +93,15 @@ def propose(self, model, state):
             new_log_probs, new_blobs = model.compute_log_prob_fn(q)
 
             # Loop over the walkers and update them accordingly.
+            try:
```
andyfaff: I would get rid of the try/except block (see below)

src/emcee/moves/red_blue.py
```diff
@@ -93,11 +93,15 @@ def propose(self, model, state):
...
+                rg_random = model.random.random
+            except AttributeError:
+                rg_random = model.random.rand
             for i, (j, f, nlp) in enumerate(
                 zip(all_inds[S1], factors, new_log_probs)
             ):
                 lnpdiff = f + nlp - state.log_prob[j]
-                if lnpdiff > np.log(model.random.rand()):
+                if lnpdiff > np.log(rg_random()):
```
andyfaff: use:
```
if lnpdiff > np.log(model.random.uniform()):
```

src/emcee/moves/red_blue.py
andyfaff: you can use `model.random.uniform`, it works with `RandomState` and `Generator`.

src/emcee/moves/mh.py
```diff
@@ -56,7 +56,11 @@ def propose(self, model, state):
 
         # Loop over the walkers and update them accordingly.
         lnpdiff = new_log_probs - state.log_prob + factors
-        accepted = np.log(model.random.rand(nwalkers)) < lnpdiff
+        try:
+            rg_random = model.random.random
+        except AttributeError:
+            rg_random = model.random.rand
+        accepted = np.log(rg_random(nwalkers)) < lnpdiff
```
andyfaff: You can use:
```
accepted = np.log(model.random.uniform(size=nwalkers)) < lnpdiff
```

src/emcee/tests/unit/test_sampler.py
```diff
@@ -319,3 +325,50 @@ def test_walkers_independent_randn_offset_longdouble(nwalkers, ndim, offset):
...
+        packaging.version.parse(np.__version__) < packaging.version.parse("1.17.0"),
+        reason="requires numpy 1.17.0 or higher",
+    )
+def test_sampler_generator():
+    nwalkers = 32
+    ndim = 3
+    nsteps = 25
+    np.random.seed(456)
+    coords = np.random.randn(nwalkers, ndim)
+    seed = np.random.default_rng()
```
andyfaff: You'll need to seed the generator:
`seed = np.random.default_rng(1)`

src/emcee/tests/unit/test_sampler.py
```diff
@@ -319,3 +325,50 @@ def test_walkers_independent_randn_offset_longdouble(nwalkers, ndim, offset):
...
+    nwalkers = 32
+    ndim = 3
+    with pytest.raises(TypeError, match="seed must be"):
+        EnsembleSampler(nwalkers, ndim, normal_log_prob, seed="bad_seed")
+
+@pytest.mark.skipif(
+        packaging.version.parse(np.__version__) < packaging.version.parse("1.17.0"),
+        reason="requires numpy 1.17.0 or higher",
+    )
+def test_sampler_generator():
```
andyfaff: this test checks that you can use a Generator to seed the Sampler, but it doesn't prove that sampling is reproducible with it. I would reduce the number of steps to 5, create an `EnsembleSampler` with a seeded `Generator` and `run_mcmc` it. Then I would then create another EnsembleSampler with a `Generator` created with the same seed, and `run_mcmc` it. THen check that the output from the two sampling runs are identical.

src/emcee/tests/unit/test_sampler.py
```diff
@@ -319,3 +325,50 @@ def test_walkers_independent_randn_offset_longdouble(nwalkers, ndim, offset):
...
+    nwalkers = 32
+    ndim = 3
+    nsteps = 25
+    np.random.seed(456)
+    coords = np.random.randn(nwalkers, ndim)
+    seed = np.random.default_rng()
+    sampler = EnsembleSampler(nwalkers, ndim, normal_log_prob, seed=seed)
+    sampler.run_mcmc(coords, nsteps)
+    assert sampler.get_chain().shape == (nsteps, nwalkers, ndim)
```
andyfaff: np.testing.assert_equal

src/emcee/tests/unit/test_sampler.py
```diff
@@ -319,3 +325,50 @@ def test_walkers_independent_randn_offset_longdouble(nwalkers, ndim, offset):
...
+    assert sampler.get_log_prob().shape == (nsteps, nwalkers)
```
andyfaff: np.testing.assert_equal

src/emcee/tests/unit/test_sampler.py
```diff
@@ -319,3 +325,50 @@ def test_walkers_independent_randn_offset_longdouble(nwalkers, ndim, offset):
...
+    for sampler in (sampler1, sampler2, sampler3, sampler4):
+        sampler.run_mcmc(coords, nsteps)
+    for k in ["get_chain", "get_log_prob"]:
+        attr1 = getattr(sampler1, k)()
+        attr2 = getattr(sampler2, k)()
+        attr3 = getattr(sampler3, k)()
+        attr4 = getattr(sampler4, k)()
+        assert not np.allclose(attr1, attr2), "inconsistent {0}".format(k)
+        assert np.allclose(attr1, attr3), "inconsistent {0}".format(k)
+        assert np.allclose(attr1, attr4), "inconsistent {0}".format(k)
```
andyfaff: np.testing.assert_allclose

src/emcee/tests/unit/test_sampler.py
```diff
@@ -319,3 +325,50 @@ def test_walkers_independent_randn_offset_longdouble(nwalkers, ndim, offset):
...
+    )
+    for sampler in (sampler1, sampler2, sampler3, sampler4):
+        sampler.run_mcmc(coords, nsteps)
+    for k in ["get_chain", "get_log_prob"]:
+        attr1 = getattr(sampler1, k)()
+        attr2 = getattr(sampler2, k)()
+        attr3 = getattr(sampler3, k)()
+        attr4 = getattr(sampler4, k)()
+        assert not np.allclose(attr1, attr2), "inconsistent {0}".format(k)
+        assert np.allclose(attr1, attr3), "inconsistent {0}".format(k)
```
andyfaff: np.testing.assert_allclose

src/emcee/ensemble.py
```diff
@@ -147,8 +158,7 @@ def __init__(
 
         # This is a random number generator that we can easily set the state
         # of without affecting the numpy-wide generator
-        self._random = np.random.mtrand.RandomState()
-        self._random.set_state(state)
+        self._check_random_state(seed, state)
```
andyfaff: `self._random = _check_random_state(seed, state)`

src/emcee/ensemble.py
```diff
@@ -178,6 +191,31 @@ def random_state(self, state):
         except:
             pass
 
+    def _check_random_state(self, seed, state):
```
andyfaff: `_check_random_state` should probably be a standalone function:

```
    def _check_random_state(seed, state=None):
        """Check seed argument and set RandomState.
        Based on scikit-learn utils/validation.py.
        """
        if isinstance(seed, (int, np.integer)) or seed is None:
            _random = np.random.mtrand.RandomState()
            if state is not None:
                _random.set_state(state)
            if seed is not None:
                _random.seed(seed)
            return _random
        elif isinstance(seed, np.random.RandomState):
            _random = seed
            if state is not None:
                _random.set_state(state)
            return _random
        else:
            try:
            # Generator is only available in numpy >= 1.17
                if isinstance(seed, np.random.Generator):
                    return seed
            except AttributeError:
                pass
            raise TypeError(
                "seed must be an int, np.random.RandomState, np.random.Generator or "
                "None of type {}".format(type(seed))
            )
```

andyfaff: I think it's only called from the `random_state` property setter, nor does it access much object state, so there's no real need for it to be an instance method.

andyfaff:
I just asked a question on scipy-dev about dealing with functions that are in Generator, but not RandomState:
The email recommended adding a function along the lines of:

```
try
    Generator = np.random.Generator
except AttributeError:
    Generator = None

def rng_integers(gen, low, high=None, ...):
    if isinstance(gen, Generator):
        return gen.integers(low, high=high, ...)
    else:
        return gen.randint(low, high=high, ...)
```

this could be added to `util`. We could write a similar one for random floating point numbers (`np.random.rand`).

OriolAbril:
@dfm @andyfaff I have done the suggested changes, thanks for all the help!

I have also modified an integration test to check everything works also with numpy Generators, would this be useful? I can extend the changes to other tests or undo them

src/emcee/ensemble.py
```diff
@@ -147,8 +158,7 @@ def __init__(
 
         # This is a random number generator that we can easily set the state
         # of without affecting the numpy-wide generator
-        self._random = np.random.mtrand.RandomState()
-        self._random.set_state(state)
+        self._random = _check_random_state(seed, state)
```
andyfaff: Does the initialisation of the `EnsembleSampler` work  if an initialized `backend` is provided, and the backend was created using a `np.random.Generator`?

src/emcee/utils.py
```diff
@@ -60,3 +60,47 @@ def sample_ellipsoid(p0, covmat, size=1):
...
+        random.set_state(state)
+        if seed is not None:
+            random.seed(seed)
+        return random
+    if isinstance(seed, np.random.RandomState):
+        return seed
+    try:
+    # Generator is only available in numpy >= 1.17
+        if isinstance(seed, np.random.Generator):
+            return seed
```
andyfaff: Could/should the state of the Generator also be set? e.g.
```
if isinstance(seed, Generator):
    if isinstance(state, dict):
        seed.bit_generator.state = state
    return seed
```
? Checking that state is a dict isn't going to solve all issues if state isn't what the bit generator is expecting, but it'll catch most things.

src/emcee/ensemble.py
```diff
@@ -164,7 +174,10 @@ def random_state(self):
         so silently.
 
         """
-        return self._random.get_state()
+        try:
```
andyfaff: Should this be:
```
if isinstance(self._random, np.random.RandomState):
    return self._random.get_state()
elif isinstance(self._random, Generator):
    return self._random.bit_generator.state
```
Obviously Generator will have to be imported from somewhere, in a similar fashion to what's done in `utils.py`.

src/emcee/ensemble.py
```diff
@@ -164,7 +174,10 @@ def random_state(self):
...
+            return self._random.get_state()
+        except AttributeError:
+            return self._random.bit_generator.state
 
     @random_state.setter  # NOQA
     def random_state(self, state):
```
andyfaff: Should we have:
```
try:
    if isinstance(self._random, np.random.RandomState):
         self._random.set_state(state)
    elif isinstance(self._random, Generator) and isinstance(state, dict):
        self._random.bit_generator.state = state
except:
    pass
```

Obviously `Generator` will have to be imported from somewhere, in a similar fashion to what's done in `utils.py`.

src/emcee/utils.py
```diff
@@ -60,3 +60,47 @@ def sample_ellipsoid(p0, covmat, size=1):
...
+    Based on scikit-learn utils/validation.py.
+    """
+    if isinstance(seed, (int, np.integer)) or seed is None:
+        random = np.random.mtrand.RandomState()
+        random.set_state(state)
+        if seed is not None:
+            random.seed(seed)
+        return random
+    if isinstance(seed, np.random.RandomState):
+        return seed
```
andyfaff: can't we also set the state here?
```
if isinstance(seed, np.random.RandomState):
    if state is not None:
         seed.set_state(state)
    return seed
```

src/emcee/ensemble.py
andyfaff: In fact if the `state` can also be set in `_check_random_state` then you could just have:
```
try:
    self._random = _check_random_state(self._random, state)
except:
    pass
```

src/emcee/ensemble.py
OriolAbril: Not sure, I'll check and get back to you, from a quick look it will probably not work

src/emcee/utils.py
OriolAbril: We could, that depends on what do we want the seed argument to do. Currently the state is only used if the seed in None where the exact same state is used, or if it is an int, in which case the state is used and reseeded. Using RandomState of Generator overwrites the stored state to use directly the passed object. 

Also, the behaviour does not depend on reusing a backend or not, maybe it should?

carlosgmartin:
@dfm What's the status of this?

dfm:
@carlosgmartin: you know, I have no idea! I think it must have gotten automatically closed when I deleted the `master` branch and renamed the default to `main`. I'll see what I can do to re-open this so that we can revive it. Thanks for catching this!

dfm:
@OriolAbril: I can't seem to change the base branch of this pull request. Are you able to do it: https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/changing-the-base-branch-of-a-pull-request

OriolAbril:
I can't either, there is no dropdown option. I can open it again but don't have much time right now to work on that, @carlosgmartin would you be interested in taking over?

carlosgmartin:
@dfm @OriolAbril Can you describe the issue in more detail? Have you tried reaching out to [GitHub Support](https://support.github.com/)?

OriolAbril:
In general, the target branch in a PR has a dropdown when editing the title of the PR as shown in the documentation linked by Dan:

image

In this case however, probably due to the original target having been deleted, neither Dan nor I have this dropdown available, so it looks like we can't reopen that PR and I would have to open a new PR to `main` with the same code.

dfm:
Thanks for checking @OriolAbril and no worries about not having time. I'll open a new pull request with your commits and we can continue the discussion over there.

Thanks all!

PR #333, opened 2019-12-30
fix norm of zero

gboehl:
If `s[i]` and `z` are the same, `norm = 0` which results in `NaNs`. The fix is trivial.

### DIFF
--- a/src/emcee/moves/de_snooker.py
+++ b/src/emcee/moves/de_snooker.py
@@ -35,11 +35,13 @@ class DESnookerMove(RedBlueMove):
         q = np.empty((Ns, ndim), dtype=np.float64)
         metropolis = np.empty(Ns, dtype=np.float64)
         for i in range(Ns):
-            w = np.array([c[j][random.randint(Nc[j])] for j in range(3)])
-            random.shuffle(w)
-            z, z1, z2 = w
-            delta = s[i] - z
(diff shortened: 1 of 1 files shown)


dfm:
Thanks for the contribution.

Unfortunately this is not a good solution for two reasons:

1. Asking a truthy question about a float is not a good idea (using something like `np.allclose` would be better).
2. If `s[i] == z` then something else is going wrong (probably an initialization issue!) because different walkers shouldn't ever have identical coordinates.

Can you post some sample code that hits this issue so that we can understand what's going on?

Thanks.

gboehl:
Thanks for looking into this! I admit that this is a quite ad-hoc fix. To your comments:

I think the truthy question here is okay as I really only wanted to capture the `s[i]==z` case. But of course you are right that this actually shouldn't happen.

Unfortunately I can't post an example. This only happend once in a large scale estimation where the initial values are drawn from the prior distribution. So the problem probably lies there :)

gboehl:
You were right, there was a bug in the sampler from the prior. Thanks for pointing me to the right place!

gboehl:
And many thanks for providing this package! It is extremely useful and very well done!

dfm:
Awesome! Glad you tracked it down. It might be worth adding a check that handles this case more gracefully, but we'll have to think a bit about how to implement that.

PR #347, opened 2020-06-25
Refactor backend classes to facilitate adding more

DavidMStraub:
Hi,

I wanted to add one or two additional backends - specifically `numpy.savez` and possibly SQLite (or generic SQLAlchemy). However when I looked at the code, the existing `Backend` and `HDFBackend` did not inherit from a common base class and it was not very transparent in my view which methods were shared/generic and which were implementation dependent. So in this PR, I just reshuffled the code - without changing anything about the API - such that implementation-independent methods are relegated to an abstract base class. Furthermore, an abstract class for file-based backends is added. In this way, it would be straightforward for me to add a `numpy.savez` backend as a next step.

Sorry I did not discuss this on the emcee-useres mailing list before opening the PR - I made the mistake to read the contributing guidelines only when it was done - but I hope you find it useful nevertheless.

### DIFF
--- a/src/emcee/backends/backend.py
+++ b/src/emcee/backends/backend.py
@@ -4,15 +4,16 @@ import numpy as np
 
 from .. import autocorr
 from ..state import State
+from .base import BackendBase
 
 __all__ = ["Backend"]
 
 
-class Backend(object):
+class Backend(BackendBase):
     """A simple default backend that stores the chain in memory"""
 
     def __init__(self, dtype=None):
-        self.initialized = False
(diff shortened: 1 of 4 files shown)


dfm:
This is awesome - thanks! I'll try to do a proper review ASAP.

DavidMStraub:
Should I rebase this?

dfm:
@DavidMStraub: Sorry about dropping the ball on this!! Yes - that would be amazing if you're up for it. Thanks for your patience.

PR #349, opened 2020-07-23
Make `emcee` more like `fortnite` 🥴

barentsen:
### Summary

This PR enables `emcee` to attract more Gen Z users by modifying the user experience to be more akin to Fortnite. 🎮

### Demo

Screenshot from 2020-07-22 19-12-35


dfm:
The builds are failing because tqdm is not a required dependency. I think that bugs/lack of tests is also consistent with the fortnite experience....

barentsen:
> dfm closed this

?!

Issue #356, opened 2020-08-18
Tests do not verify emcee works without dev dependencies


aarchiba:
**General information:**

- emcee version: 3.0.2
- platform: Linux
- installation method (pip/conda/source/other?):

**Problem description:**

If `emcee` is supposed to work with only it's declared dependencies, one should be able to simply install it and `pytest` and run `pytest` and all relevant tests should pass. If not, how will we notice if we unintentionally introduce, say, a dependency on `h5py` in the mainline code?

```sh
$ mkvirtualenv emcee
$ pip install -e .
$ pip install pytest
$ pytest
```

### Expected behavior:

All tests pass, are skipped, or are xfailed.

### Actual behavior:

Many test files cannot even be imported without scipy.

PR #354, opened 2020-08-18
Fix long double test failures with HDF5 and ppc64el

aarchiba:
Detects whether HDF5 supports long doubles and `xfail`s tests if it doesn't.

Closes #353

PR #358, opened 2020-08-18
Tox tests without scipy and works

aarchiba:
Set up `tox` to run tests without the developer requirements installed; also fix the code so that it works if `scipy` is not installed. Doesn't work without `h5py`; for that see #354 ,

Partially addresses #356 .


.travis.yml
```diff
@@ -1,24 +1,34 @@
...
+    - python: "3.8"
+      env: NUMPY_VERSION=default
+      install:
+          - travis_retry pip install --upgrade pip
+          - travis_retry pip install --upgrade setuptools
+          - travis_retry pip install pytest pytest-cov coveralls setuptools-scm
+          - travis_retry pip install h5py
+          - pip install -vv .
+          - python -c 'import emcee;print(emcee.__version__)'
+          - travis_retry python setup.py develop
```
dfm, line 31: Instead of running `setup.py develop` here, it would be better to have `pip install -e -vv .` above, I think!

PR #362, opened 2020-08-23
Add HDF5 compression

### LINKED ISSUE #338: Enable HDF5 compression in the backend
**General information:**

- emcee version: 3.0.2
- platform: (Arch) Linux
- installation method (pip/conda/source/other?): AUR repository

**Problem description:**
It would be nice to be able to enable compression when using HDF5 files via the emcee backends. A quick skim of the documentation and `backends/hdf.py` file suggests this isn't currently possible. Perhaps the h5py `compression=` argument could be exposed in the `emcee.backends.HDFBackend` class?

aarchiba:
Allow users to enable compression on HDF5 backends.

Compression is applied uniformly to all datasets; one might expect different compression for `accepted` than `chain` or `blobs` but the interface doesn't need additional complications. If `reset` is used on an existing HDF5 file the newly created datasets will have compression enabled or not as specified during backend creation.

Closes #338

### DIFF
--- a/src/emcee/backends/hdf.py
+++ b/src/emcee/backends/hdf.py
@@ -22,9 +22,9 @@ except ImportError:
 def does_hdf5_support_longdouble():
     if h5py is None:
         return False
-    with NamedTemporaryFile(prefix="emcee-temporary-hdf5",
-                            suffix=".hdf5",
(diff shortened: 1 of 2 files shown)


dfm:
Awesome!!

PR #363, opened 2020-08-23
Solidify blobs dtype inference

### LINKED ISSUE #361: Emits numpy deprecation warning about ragged sequences
**General information:**

- emcee version: 3.0.3.dev5+g87c4c85.d20200818
- platform: Linux
- installation method (pip/conda/source/other?): pip

**Problem description:**

Currently the way `emcee` packs blobs into its storage array risks incompatibility with future versions of `numpy` - or at least it triggers a DeprecationWarning. The basic problem is that `src/emcee/ensemble.py` tries to figure out the right dtype with `atleast_1d`:

```python
            # Get the blobs dtype
            if self.blobs_dtype is not None:
                dt = self.blobs_dtype
            else:
                try:
                    dt = np.atleast_1d(blob[0]).dtype
                except ValueError:
                    dt = np.dtype("object")
            blob = np.array(blob, dtype=dt)
```

The current behaviour will let users return lists/tuples of numbers and have them converted to extra dimensions on a numeric blobs array; this will come to grief if the lengths of those lists/tuples ever change or if the types change. But in the vast majority of cases the user probably wants numeric arrays rather than object arrays even if they just do `return 2, 3`. That said strings will make it very unhappy. Correct behaviour is unclear.

### Expected behavior:

`emcee` test suite completes without warnings.

### Actual behavior:

`emcee` test suite emits `numpy.VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray`

### Minimal example:

```
pytest -Werror src/emcee/tests/unit/test_blobs.py
```

aarchiba:
Reduce the chance of user surprise due to blob dtype inference - strings are object dtypes, object dtypes don't lead to backend weirdness, ragged blob dtypes raise `numpy.VisibleDeprecationWarning` unless `blobs_dtype` is specified. Also fixes problems where `grow` used to re-infer blob dtypes. 

Closes #361

### DIFF
--- a/src/emcee/tests/unit/test_blobs.py
+++ b/src/emcee/tests/unit/test_blobs.py
@@ -51,7 +51,7 @@ def test_blob_shape(backend, blob_spec):
 
         if ragged:
             with warnings.catch_warnings():
-                warnings.simplefilter("ignore", np.VisibleDeprecationWarning)
+                warnings.simplefilter("ignore", DeprecationWarning)
                 sampler.run_mcmc(coords, nsteps)
         else:
             sampler.run_mcmc(coords, nsteps)
--- a/src/emcee/tests/unit/test_blobs.py
+++ b/src/emcee/tests/unit/test_blobs.py
@@ -22,8 +22,9 @@ class BlobLogProb(object):
 @pytest.mark.parametrize(
     "blob_spec",
     [
-        (True, False, (5, 3), lambda x: np.random.randn(5, 3)),
         (True, False, 5, lambda x: np.random.randn(5)),
+        (True, False, (5, 3), lambda x: np.random.randn(5, 3)),
+        (True, False, (5, 3), lambda x: np.random.randn(1, 5, 1, 3, 1)),
         (True, False, 0, lambda x: np.random.randn()),
         (False, True, 2, lambda x: (1.0, np.random.randn(3))),
         (False, False, 0, lambda x: "face"),
--- a/src/emcee/tests/unit/test_blobs.py
+++ b/src/emcee/tests/unit/test_blobs.py
@@ -50,7 +50,8 @@ def test_blob_shape(backend, blob_spec):
         nsteps = 10
 
         if ragged:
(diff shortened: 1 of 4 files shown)


src/emcee/tests/unit/test_blobs.py
```diff
@@ -20,32 +22,73 @@ def __call__(self, params):
...
     "blob_spec",
     [
-        (True, 5, lambda x: np.random.randn(5)),
-        (True, 0, lambda x: np.random.randn()),
-        (False, 2, lambda x: (1.0, np.random.randn(3))),
-        (False, 0, lambda x: "face"),
-        (False, 2, lambda x: (np.random.randn(5), "face")),
+        (True, False, (5, 3), lambda x: np.random.randn(5, 3)),
+        (True, False, 5, lambda x: np.random.randn(5)),
+        (True, False, 0, lambda x: np.random.randn()),
```
aarchiba, line 27: Any dimensions of length 1 get squeezed out, so `randn(1)` would produce the same shape.

dfm:
This looks great!

Like you suggested in #361, perhaps we should catch the cases where this deprecation hits and a user hasn't provided a `blobs_dtype` and provide a recommendation? Happy to merge this and open a separate issue if you think that's best.

Thanks!!

aarchiba:
https://numpy.org/neps/nep-0034-infer-dtype-is-object.html specifies the deprecation process, so the same change should allow transforming the ValueError into a more informative one as well, for users with a whiz-bang future version of numpy that does that.

aarchiba:
Ready for review.

dfm:
Sorry I've been slow to respond - this week has been crazy! I'll get back here in the next few days.

dfm:
Sorry about the delay! I think this looks great and I'm happy to merge. Thanks for this! I'll start a new pull request thread to draft a new release.

Issue #369, opened 2020-12-19
emcee.EnsembleSampler.sample with infinite iterations


carlosgmartin:
**General information:**

- emcee version: 3.0.2
- platform: Python 3.8.5
- installation method (pip/conda/source/other?): `conda install -c conda-forge emcee`

**Problem description:**

It would be nice to be able to use [`emcee.EnsembleSampler.sample`](https://emcee.readthedocs.io/en/stable/user/sampler/#emcee.EnsembleSampler.sample) with infinite `iterations`, producing an unending stream of samples. This frees the caller from having to specify the number of iterations ahead of time, and allows them to take as many samples as they need *post facto*. I find myself using this pattern often, so I think it would be good from a UI standpoint.

dfm:
This was possible in previous versions of emcee, but while I agree that it's elegant in theory, no one ever actually interfaced with the code that way and it's not such a good idea with the current version because of how the chains are stored. I tested an implementation where the backend memory wasn't pre-allocated and there was a huge computational overhead every time you need to resize the array. All that to say: if you *actually* want an infinite sampler, you should figure out what block size works for your problem and then put the `sampler.sample` calls (with that number of iterations) inside a `while True`. It's easy to restart where you left off (just pass an initial state of `None`). 

Take a look at this tutorial that gives a few examples the would be building blocks: https://emcee.readthedocs.io/en/stable/tutorials/monitor/

carlosgmartin:
@dfm Does this apply with `store=False`?

dfm:
You won't get any resizing overhead when `store=False`, true, but what's your use case where you want to run an infinite number of steps without saving them anywhere? And why is infinity better than a large integer? The default is `store=True` because, in my experience, that's how most people use the sampler, but I'm interested to understand why you think that something else would be a better UI!

carlosgmartin:
It's a bit hard to explain, but I'm estimating something of the form f(E<sub>X ~ p</sub> g(X)) by using emcee to take samples from p, and eventually stopping the process once a certain condition on h(E<sub>X ~ p</sub> i(X)) is met. I don't know ahead of time when this condition will be met. I also plot progress along the way.

I can work around it, of course (using `run_mcmc` as you suggested), but I've been using this pattern often in my research, so I thought it might be nice to have that option.

dfm:
I see - so you just track the value of g(X) at each step and you don't need the whole chain?

[...]

PR #371, opened 2020-12-20
Added infinite iterations option to EnsembleSampler.sample

carlosgmartin:
Added option to use `iterations=None` in [`EnsembleSampler.sample`](https://emcee.readthedocs.io/en/stable/user/sampler/#emcee.EnsembleSampler.sample), as discussed in #369. This produces an infinite stream of samples. Requires `store=False`. Note that tqdm can handle `total=None`. Example:

```python3
import numpy as np
from emcee import EnsembleSampler

def logp(x):
    return np.where(((0 <= x) & (x <= 1)).all(-1), 0, -np.inf)

x = np.random.rand(10, 2)

for state in EnsembleSampler(*x.shape, logp).sample(x, progress=True, iterations=None, store=False):
    pass
```

### DIFF
--- a/src/emcee/ensemble.py
+++ b/src/emcee/ensemble.py
@@ -247,7 +247,7 @@ class EnsembleSampler(object):
         :class:`State` of the ensemble.
 
         """
-        if iterations is None and not store is False:
+        if iterations is None and store:
             raise ValueError("'store' must be False when 'iterations' is None")
         # Interpret the input as a walker state and check the dimensions.
         state = State(initial_state, copy=True)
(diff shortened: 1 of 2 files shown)


dfm:
Thanks! This looks great. Can you add a test or two? We should test that the exception gets raised when expected and that the iterations work as expected when set to `None`.

A good place to put it is here: https://github.com/dfm/emcee/blob/main/src/emcee/tests/unit/test_sampler.py

carlosgmartin:
Added.

dfm:
Looks good - thanks!

PR #376, opened 2021-01-16
Add seed argument to EnsembleSampler

dfm:
This was previously opened as #332, but because of branch renaming issues we had to reopen it. This is @OriolAbril's work, but @carlosgmartin might also be interested. I'll try to summarize where we stand from the previous thread shortly, but I think we were just about there. Anything in particular that was missing?

### DIFF
--- a/src/emcee/ensemble.py
+++ b/src/emcee/ensemble.py
@@ -159,24 +159,7 @@ class EnsembleSampler(object):
 
         # This is a random number generator that we can easily set the state
         # of without affecting the numpy-wide generator
-        if isinstance(seed, int) or seed is None:
-            self._random = np.random.mtrand.RandomState()
-            self._random.set_state(state)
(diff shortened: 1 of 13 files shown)


dfm:
The tests are failing because we need to change

```python
import packaging
```

to

```python
import packaging.version
```

dfm:
@OriolAbril: If you check the "allow edits from maintainers" checkbox, then I can take over from here.

OriolAbril:
I don't have the option too, but I can give both you and @carlosgmartin access to my fork.

andyfaff:
I should point out [SPEC-007](https://scientific-python.org/specs/spec-0007/), which is trying to normalise how random numbers are using within the scientific ecosystem. The main items are converging on the use of an `rng` keyword, and a move to using `Generator` instead of of `RandomState`

PR #380, opened 2021-02-18
Fix np.float deprecation

JBorrow:
`np.float` is deprecated as of numpy 1.20. https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations.


dfm:
Thanks!

PR #386, opened 2021-04-24
Support for named parameters

### LINKED ISSUE #385: [ENH] support for dictionaries and/or namedtuples
## Proposal

The sampler API shares a lot of features with `scipy.optimize.minimize` -- given a model that can be typed as `Callable(List,...)` keep proposing samples to improve the model. This is fine for many problems, but one can imagine a scenario where it would be useful for model developers to handle samples with more than just a `List`. Handling parameters in a `List` means that writing a posterior *requires defining parameters positionally*. In contrast, in packages like pytorch one can refer to model parameters by name (eg. a `torch.nn.Linear` has a `weight` and `bias` named parameters).

This issue proposes enhancing the emcee API to allow for dictionary and/or namedtuple parameter constructs.  Looking at the API for the `EnsembleSampler`, this would require an additional constructor argument, since the sampler does not inspect the state of the `log_prob_fn`.

Here is a minimum proposed implementation for supporting a dictionary:

```python
class EnsembleSampler(object):
    def __init__(
        self,
        nwalkers: int,
        ndim: int,
        log_prob_fn: Callable(Union[List, Dict]),
        parameter_names: Optional[List[str]] = None,
        ...
    ):
        ...

        # Save the names
        self.parameter_names = parameter_names
        self.name_parameters: bool = parameter_names is not None
        if self.name_parameters:
            assert len(parameter_names) == ndim, f"names must match dimensions"
        ...

    def sample(...):
        ...
         with get_progress_bar(progress, total) as pbar:
            i = 0
            for _ in count() if iterations is None else range(iterations):
                for _ in range(yield_step):
                    ...

                    # If our parameters are named, return them in a named format
                    if self.name_parameters:
                        state = {key: value for key, value in zip(self.parameter_names, state)}

                    yield state
```

This allows for a lot more flexibility downstream for model developers. As a simple example, consider a model composed of two parts, that each can be updated in a way that doesn't affect the higher level model:

```python
class PartA:
    __slots__ = ["param_a"]
    param_a: float

    def __call__(self):
        ...

class PartB:
    __slots__ = ["param_b"]
    param_b: float

    def __call__(self):
        ...

class MyModel:
    def __init__(self, a = PartA(), b = PartB()):
        self.a = a
        self.b = b

    def log_likelihood_fn(self, params: Dict[str, float]):
        # Update the composed parts of the model
        for key in self.a.__slots__:
            setattr(self.a, params[key])

        for key in self.b.__slots__:
            setattr(self.b, params[key])

        # Final likelihood uses the results of A and B, but we don't
        # care how they are implemented
        return f(self.a(), self.b())
```

If this sounds like an attractive feature I would be happy to work on a PR.

tmcclintock:
This PR closes #385 by giving support for named parameters in the `EnsembleSampler`.

Specifically this PR:
- allows a user to pass in a list of names (types as `str` but really can be anything) for the parameters to the constructor for `EnsembleSampler`. Checks are made to ensure that the number of unique names are `ndim`
- In the `EnsembleSampler.compute_log_prob` method, these names are used to reshape the parameters passed in to the method (having previously been generated by the `Move`) so that instead of an `ndarray` of shape `(nwalkers, ndim)` it becomes a list of length `nwalkers` of dictionaries of length `ndim`, with each key-value pair being the names and parameters
- A helper function was written to perform the `ndarray` -> `List[Dict[str, np.number]]` conversion, so that it could be tested separately from the `compute_log_prob` method (higher testable surface area)
- test of this functionality were placed in the newly created `src/emcee/tests/unit/test_ensemble.py` file. I tested the helper method, `compute_log_prob`, and `run_mcmc`
- very small bits of cleanup in `ensemble.py` to be more PEP8 compliant/more pythonic

As discussed in #385, this PR makes no assumptions about the kinds of likelihood functions or underlying models being utilized. It does not take gradients of the likelihood, nor does it break the API (frontend or backend) in any way. I believe the unit test in in the `TestNamedParametes` tests make it clear why this is useful. Positional bugs will be a thing of the past!

Support for named parameters could go a bit further. For example:
1. I didn't allow `vectorize=True` when naming the parameters. I guess the right thing to allow for would be if `vectorize=True` then we produce **one** dictionary where the keys are the parameter names and the values are 1D arrays of each parameter.
2. I did not implement support for parameter names in the `PTSampler`. IMO doing this would be "double work" and from a software perspective it would make more sense (IMO) to make an abstract `Sampler` class that handled the names, and that both the `EnsembleSampler` and `PTSampler` would subclass. But, this is obviously a non-trivial refactor (that I would be happy to contribute to).

Thanks @dfm for the discussion on #385. I look forward to your review.

### DIFF
--- a/src/emcee/ensemble.py
+++ b/src/emcee/ensemble.py
@@ -168,23 +168,28 @@ class EnsembleSampler(object):
         self.params_are_named: bool = parameter_names is not None
         self.parameter_names: Optional[List[str]] = parameter_names
         if self.params_are_named:
+            assert isinstance(parameter_names, (list, dict))
+
             # Don't support vectorizing yet
             msg = "named parameters with vectorization unsupported for now"
             assert not self.vectorize, msg
-            
-            # Check for all named
-            msg = "name all parameters or set `parameter_names` to `None`"
-            assert len(parameter_names) == ndim, msg
 
-            # Check for duplicate names
-            dupes = set()
-            uniq = []
-            for name in parameter_names:
-                if name not in dupes:
-                    uniq.append(name)
-                    dupes.add(name)
-            msg = f"duplicate paramters: {dupes}"
-            assert len(uniq) == len(parameter_names), msg
+            if isinstance(parameter_names, list):
+                # Check for all named
+                msg = "name all parameters or set `parameter_names` to `None`"
+                assert len(parameter_names) == ndim, msg
+
+                # Check for duplicate names
+                dupes = set()
+                uniq = []
+                for name in parameter_names:
+                    if name not in dupes:
+                        uniq.append(name)
+                        dupes.add(name)
+                msg = f"duplicate paramters: {dupes}"
+                assert len(uniq) == len(parameter_names), msg
+            else:
+                pass  # TODO
 
     @property
     def random_state(self):
--- a/src/emcee/ensemble.py
+++ b/src/emcee/ensemble.py
@@ -2,7 +2,7 @@
 
 import warnings
 
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Union
 
 import numpy as np
 from itertools import count
@@ -63,6 +63,10 @@ class EnsembleSampler(object):
             to accept a list of position vectors instead of just one. Note
             that ``pool`` will be ignored if this is ``True``.
             (default: ``False``)
+        parameter_names (Optional[Union[List[str], Dict[str, List[int]]]]):
+            names of individual parameters or groups of parameters. If
+            specified, the ``log_prob_fn`` will recieve a dictionary of
+            parameters, rather than a ``np.ndarray``.
 
     """
 
@@ -78,7 +82,7 @@ class EnsembleSampler(object):
         backend=None,
         vectorize=False,
         blobs_dtype=None,
-        parameter_names: Optional[List[str]] = None,
+        parameter_names: Optional[Union[Dict[str, int], List[str]]] = None,
         # Deprecated...
         a=None,
         postargs=None,
--- a/src/emcee/ensemble.py
+++ b/src/emcee/ensemble.py
@@ -1,11 +1,10 @@
 # -*- coding: utf-8 -*-
 
 import warnings
-
+from itertools import count
 from typing import Dict, List, Optional, Union
 
 import numpy as np
-from itertools import count
 
 from .backends import Backend
 from .model import Model
--- a/src/emcee/ensemble.py
+++ b/src/emcee/ensemble.py
@@ -2,6 +2,8 @@
 
 import warnings
 
+from typing import Dict, List, Optional
+
 import numpy as np
 from itertools import count
 
@@ -76,6 +78,7 @@ class EnsembleSampler(object):
         backend=None,
         vectorize=False,
         blobs_dtype=None,
+        parameter_names: Optional[List[str]] = None,
         # Deprecated...
         a=None,
         postargs=None,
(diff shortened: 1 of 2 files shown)


dfm:
Thanks @tmcclintock: This looks like a good start!

I'll take a closer look ASAP, but I think I'm overall very happy with this approach. I think the one thing I would like to have would be to allow each parameter to have an arbitrary `shape`. This will complicate the logic a bit, but I think it would make this more impactful.

dfm:
Also: `scipy` needs to be an optional dependency even for the tests.

tmcclintock:
@dfm thanks for the quick response.

RE each parameter have an arbitrary `shape` - Going by [this docstring](https://github.com/dfm/emcee/blob/main/src/emcee/ensemble.py#L219) I was under the impression that each parameter was a scalar. Are you saying you want to change the signature to `EnsembleSampler.sample` such that `initial_state` could be `(List[Union[np.number, np.ndarray]])` (i.e. `params[0]` was a scalar but `params[1]` could be an array)?  If not, can you please clarify the pattern you are requesting?

RE `scipy` in all tests - can you elaborate? Do you mean just `import scipy` at the top? That should not be necessary unless there is some magic in the github actions you set up.

dfm:
> RE each parameter have an arbitrary shape - Going by this docstring I was under the impression that each parameter was a scalar. Are you saying you want to change the signature to EnsembleSampler.sample such that initial_state could be (List[Union[np.number, np.ndarray]]) (i.e. params[0] was a scalar but params[1] could be an array)? If not, can you please clarify the pattern you are requesting?

I'm not suggesting changing the syntax, I'm just saying that if you're adding names to elements of the vector space, there's no reason why there needs to be one name per dimension. My parameters are often multivariate, and it would be annoying to manually add `param_0`, `param_1`, ... as names manually, vs. calling it `param` and providing a shape. Right now, you're requiring that a name maps to a scalar and that seems artificial if we're adding this logic anyways!

If this isn't clear, I'll try to put together a little demo this week.

tmcclintock:
I see, so for instance in a GMM you would essentially like to do:
```python
params["weights"]
# (2,)  # component weights
params["means"].shape
# (2,)
params["cov"].shape
# (3,)  # independent elements of the cov matrix
```
Correct?

For an MVP this is simple -- if we restrict keywords to be either single values or consecutive numbers in the parameter array then this is easy. If instead we wanted to allow a keyword to map to an arbitrary set of indices in the params array, this is tricky (like if `params["a"]` mapped to the first, fifth, and tenth parameters in the params array, this may be a tricky pattern to support in general).

Are you OK with allowing only consecutive parameters?

dfm:
Yes this sounds about right and I would be totally happy with requiring contiguous values!

tmcclintock:
@dfm I have updated the PR. Now the following pattern is supported:
```python
def ln_prob(params):
    mean1, mean2 = params["means"]
    var1, var2 = params["vars"]
    constant = params["constant"]
    ...


sampler = EnsembleSampler(..., parameter_names = {"means": [0, 1], "vars": [2, 3], "constant": 4})
```

That is, `paramter_names` can be any of:
- a list of names `["a", "b", "c"]`
- a dictionary where the values can be integer positions of parameters or list of integers `{"a": [1, 3, 5], "b": 0, "c": [2, 4]}`

So, I was able to support non-contiguous parameter indexes.

Let me know if this all looks good. If you'd like docs work done I can either do that here or in another PR.

tmcclintock:
@dfm can you push the button on the workflow? I'd like to tackle any problems that might occur ahead of your review.

dfm:
@tmcclintock: Weird - I thought I already had. Maybe I have to do it each time?

dfm:
Don't worry about the coverage failure.

dfm:
@tmcclintock: Sorry about the delay on this! I think this looks awesome so I've merged and we can address issues as they come up in other threads. If you'd be up for adding a tutorial about this, that would be awesome. Want to start another PR whenever you have a chance?

Thanks again for this!

cluhmann:
Sorry to revive this, but did the tutorial mentioned ever get put together?  If so, could you point me to it?  I can probably figure it out on my own, but I figured I would check.  I'm super excited about this feature.  Thanks!

tmcclintock:
@cluhmann I have not written the tutorial. My bad :/. You can see an example of how to set up a function with named parameters in the unit test found in `src/emcee/tests/unit/test_ensemble.py`.

PR #395, opened 2021-07-12
Hdf5 swm rmode

Thalos12:
The SWMR feature of HDF5 has been implemented in the HDFBackend and is activated setting the `swmr` argument to `True`. The process reading the file should not only use the SWMR mode but also set the environment variable `HDF5_USE_FILE_LOCKING="FALSE"`.

Let me know if I should change anything I have added.

### DIFF
--- a/docs/tutorials/swmr.ipynb
+++ b/docs/tutorials/swmr.ipynb
@@ -1,0 +1,480 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "(swmr)=\n",
+    "\n",
+    "# Reading an HDF5 file while it is being written\n",
+    "\n",
+    "This tutorial will show how it is possible to read from an HDF5 file while it is being written by the `HDFBackend` without incurring into crashes of the process that is currently writing to the file.\n",
+    "\n",
+    "By default, an HDF5 file can be either read or written and it is not possible to do both at the same time. Trying to do so anyway might lead to crashes of the `emcee` process because the information in the file is not properly synchronized.\n",
+    "\n",
+    "**Important note**: the relevance of this effect depends on how much time is spent on writing relative to the time taken to perform a step. If saving takes a relatively small time, it is less likely that you will encounter crashes of the emcee process."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%config InlineBackend.figure_format = \"retina\"\n",
+    "\n",
+    "from matplotlib import rcParams\n",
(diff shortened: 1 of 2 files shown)


Thalos12:
I can see that the build of the documentation failed, but in the details I can only see a warning about the swmr.ipynb not being in any toctree. I am not familiar with sphinx, therefore I do not know how to fix this error...

Thalos12:
Nevermind, it was easy enough that I finally figured it out.

dfm:
@Thalos12: Thanks for this!! I'll need a couple of days to get to this, but as a zeroth order comment: how hard do you think it would be to add a unit test for this feature? Also it would be good to put the version requirements for h5py somewhere - what were they again?

Thalos12:
@dfm: Regarding the unit test, I have to say that I am not very experienced in writing it, but I will try my best nevertheless using those already present as examples. For the version requirements of h5py, there are none (explicitly), the requirements are on the underlying HDF5 library (>=1.10), and internally h5py already checks them, so i chose not to add duplicate code. I can mention these requirements in the tutorial notebook, if you think it would be a good idea! I can also add them to the HDFBackend docstring.

dfm:
Awesome! Yes - please add some comments about versions to the docstring and tutorial notebook.

For the unit test: no stress! I think it'll be tricky to write exactly the one we want (we'll probably need to launch multiple processes) so I'm happy to do that as I'm reviewing. Do you want to write one that just checks that the backend at least works with swm mode enabled?

Thalos12:
Thank you, I will add the requirements in both places then! Also, I suppose I should be able to make the test for the backend, I'll add a commit as soon as I have implemented it.

znicholls:
@dfm and @Thalos12 is there anything that can be done to help this? This feature looks super useful so I'd be glad to pitch in.

Thalos12:
Hi @znicholls, while making tests I discovered that the change I proposed did not work as I wanted. Going by memory, the problem is that the output hdf5 file is opened (for writing) and closed at every step of the chain ([see emcee/backends/hdf.py](https://github.com/dfm/emcee/blob/48a583b9dcf2232e2499f4ba60cf3a182830ec01/src/emcee/backends/hdf.py#L214)). Consequently, crashes can still occur if a reader opens the file just before the writer, regardless of SWMR.
I could not find a way to fix this at the time, and I forgot to close the issue (@dfm: apologies, I should have done that).

To make SWMR works, it would be necessary to modify HDFBackend to keep the file open, if at all possible, and also implement some mechanism to safely close it when needed (chain ends, exception is raised, ...).
@dfm: please correct me if I'm wrong

znicholls:
Huh that's a shame. When I read your notebook, it seemed like the reader would sometimes crash, but the writer would never crash which still seemed like a win to me and something worth including. Maybe I misunderstood

PR #399, opened 2021-07-30
Fix documentation example of blob returns for case when ll or lp is not finite

adrn:
Continuing my [long history of contributing groundbreaking / absolutely critical bug fixes to emcee](https://github.com/dfm/emcee/commits?author=adrn), here's a small documentation fix I noticed with @katiechambe! The docs example never actually hits the cases where `ll` or `lp` is infinite or nan because the functions are simple, but we copy-pasted and used in a case where we return 4 blob elements with a likelihood that sometimes goes infinite and hit a numpy shape error when returning just a single value for the blobs.


dfm:
Thanks!!

PR #401, opened 2021-08-06
Added functionality to give a description in the progress bar

arm61:
`tqdm` allows a description to be given to the progress bar, I find this is helpful in packages where multiple different progress bars are used. This PR allows a description to be added to a `emcee` progress bar based on a string passed to to the `sample` function.

### DIFF
--- a/src/emcee/ensemble.py
+++ b/src/emcee/ensemble.py
@@ -261,6 +261,7 @@ class EnsembleSampler(object):
         thin=None,
         store=True,
         progress=False,
+        progress_desc=None,
     ):
         """Advance the chain as a generator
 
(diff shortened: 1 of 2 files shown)


dfm:
Thanks! What about adding a `progress_kwargs` dictionary or something to allow other keyword arguments beyond just `desc`?

arm61:
That is a cleaner way. I have put it in like this, happy to know your thoughts.

dfm:
Looks good to me! Thanks!!

arm61:
@dfm could I be a pest and ask for a pypi release with this feature? I can't get my package onto pypi currently as it requires this feature and pypi doesn't like direct dependencies to git repos apparently.

dfm:
Absolutely! I'll do that now.

dfm:
v3.1.1 should be on PyPI shortly (as soon as [this](https://github.com/dfm/emcee/actions/runs/1159207067) finishes running)

arm61:
Amazing! Thanks for your help 🙏🏻

PR #432, opened 2022-07-27
Update link to Goodman & Weare 2010

covertg:
Hi,

Relatively new user here, and I've been noticing that the URL to Goodman & Weare (2010) on the front page of the docs is broken. Wanted to help and noticed it's an easy fix, so hopefully this PR points the link to the right place now.

Let me know if there's anything else I should change!


dfm:
Thanks!!

PR #436, opened 2022-09-13
Pass on random generator to KDE resample method

hendriko373:
Currently, the random generator in the `get_proposal` method for the `KDEMove` is not used when taking samples from the KDE. This is problematic since we cannot run emcee in a reproducible way. This PR fixes this by passing on the random generator to the KDE resampling method, so that sampling is reproducible.

### DIFF
--- a/src/emcee/moves/kde.py
+++ b/src/emcee/moves/kde.py
@@ -38,6 +38,6 @@ class KDEMove(RedBlueMove):
     def get_proposal(self, s, c, random):
         c = np.concatenate(c, axis=0)
         kde = gaussian_kde(c.T, bw_method=self.bw_method)
-        q = kde.resample(len(s))
+        q = kde.resample(len(s), random)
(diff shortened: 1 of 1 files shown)


hendriko373:
The [test ](https://github.com/dfm/emcee/blob/37dc02e2350a987f092c2d5325cc324615853273/src/emcee/tests/integration/test_longdouble.py#L21) `test_longdouble_actually_needed` sometimes fails: the condition 
```python
assert np.abs(np.mean(samples) - mjd) < 10 * sigma / np.sqrt(len(samples))
```
fails for certain np.random states. 

What is the preferred way to solve this?

- Seeding the random generator?
- Relaxing the condition?
- ...

dfm:
Thanks!

hendriko373:
You're welcome. 

Would it be possible to get a new version released including this fix?

dfm:
I'm going to wait until we get that test failure figured out (#438) then definitely!

Issue #337, opened 2020-03-11
Documentation for integrated_time still specifies discontinued ``axis`` keyword argument


ljschumacher:
**General information:**

- emcee version: 3.0.2
- platform: MacOS X
- installation method (pip/conda/source/other?): pip

**Problem description:**
The docstring for the integrated_time function still mentions the argument 'axis', which does not seem to be implemented anymore. It would be great if the axis argument could be specified, to avoid having to swap axes before passing to this function.

See lines 47 onwards in autocorr.py

```def integrated_time(x, c=5, tol=50, quiet=False):
    """Estimate the integrated autocorrelation time of a time series.

    This estimate uses the iterative procedure described on page 16 of
    `Sokal's notes <http://www.stat.unc.edu/faculty/cji/Sokal.pdf>`_ to
    determine a reasonable window size.

    Args:
        x: The time series. If multidimensional, set the time axis using the
            ``axis`` keyword argument and the function will be computed for
            every other axis.
```

warrickball:
I just ran into this myself. I was poking around in the hope of possibly open a PR but I realised that I don't understand the autocorrelation well enough to implement something myself. I'm happy to just remove any reference to the `axis` argument. I don't mind looping over dimensions and walkers myself.

On a related note, though, I noticed that `integrated_time` seems to expect the time series in a different order to the default chain layout. Specifically, [line 85 of `autocorr.py`](https://github.com/dfm/emcee/blob/main/src/emcee/autocorr.py#L85) unpacks the dimensions as
```python
    n_t, n_w, n_d = x.shape
```
Assuming `n_t`, `n_w` and `n_d` are the numbers of steps, walkers and dimensions, respectively, then the default chain structure I get is `(n_w, n_t, n_d)`. I think this is the axis-swapping that @ljschumacher referred to above.

dfm:
The order is actually correct if you use the (preferred) `get_chain` method instead of the `chain` property. In the back end, the chain is stored as `n_t, n_w, n_d` as expected, but the `chain` property has the "wrong" shape for backwards compatibility:

https://github.com/dfm/emcee/blob/f0489d74250e2034d76c3a812b9afae74c9ca191/src/emcee/ensemble.py#L489

I did at one point try re-implementing the `axis` feature, but ran into some bookkeeping issues that I don't totally remember... I'd be happy to review a PR that implements this feature, but perhaps in the meantime I should just fix that docstring :D

PR #472, opened 2023-07-04
Fix integrated_time docstring

tilmantroester:
This PR addresses issue #337 by removing references to the `axis` keyword in the `integrated_time` docstring. It also adds a `has_walkers` keyword argument to make the routine more intuitive to use for cases where the time series does not contain multiple walkers (e.g. a single MH chain). It defaults to the old behaviour.

While writing a unit test I stumbled upon this test:
https://github.com/dfm/emcee/blob/main/src/emcee/tests/unit/test_autocorr.py#L38-L48

Is this doing what it is supposed to? With the current API, the 2d `xs` is interpreted as the having 2 walkers and 1 parameter. So `integrated_time` returns a scalar. Running `integrated_time` for each of the entries separately gives an array, however. 
I think I fixed this in 40904efd99e56117c1e52af0eb9b619713e93acd but do let me know if I misunderstood what the test was supposed to be testing.

### DIFF
--- a/src/emcee/autocorr.py
+++ b/src/emcee/autocorr.py
@@ -46,7 +46,7 @@ def auto_window(taus, c):
     return len(taus) - 1
 
 
-def integrated_time(x, c=5, tol=50, quiet=False):
+def integrated_time(x, c=5, tol=50, quiet=False, has_walkers=True):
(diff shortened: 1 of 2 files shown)


dfm:
Thanks!!

PR #480, opened 2023-09-30
Fixed DEMove Implementation

StarostinV:
# Problem:

- I observed that the current implementation of DEMove, although stated to follow Nelson et al. 2013, does not actually align with the paper's approach. The existing implementation lacked affine-invariance, which is fixed in Nelson's paper.
- There was potential to optimize performance via vectorization.

# Changes

Changes are only made in the implementation of DEMove in de.py.

- **Alignment with Nelson et al. 2013**:
Fixed the discrepancy between the stated and actual approaches. The normal noise $\epsilon \sim N(0, \sigma^2)$ used to be added to the proposal as per $\gamma_0 (x_i - x_j) + \epsilon$. In this case, $\sigma$ should be tuned for each dimension, which confronts the idea of affine-invariance. Now it defines the distribution over gamma: $\gamma = \gamma_0 (1 + \epsilon)$, which follows Nelson et al. 2013 and ensures affine-invariance. 

- **Vectorization for improved performance:**
Removed a loop in favor of vectorized operations. Benchmarks showed a speed boost ranging from 5-10 times depending on the number of walkers.

- **Simplified walker splitting:**
The new implementation requires splitting the walker population into two sets as opposed to the previous three-set approach.

### DIFF
--- a/src/emcee/moves/de.py
+++ b/src/emcee/moves/de.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from functools import lru_cache
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


dfm:
Thanks! This looks good to me!

Issue #483, opened 2023-10-15
Problem using 4-byte floats with EnsembleSampler


emprice:
## General information

- emcee version: 3.14
- platform: Ubuntu Linux, Python 3.11.5
- installation method: pip

## Problem description

For context, the reason I'm doing this at all is because I wanted to benchmark some code that uses a GPU likelihood function, in vectorized mode, to see if it gives a performance boost. 4-byte floats are better-supported on GPUs, and they tend to perform better.

### Expected behavior

When the input state is given as a NumPy ndarray with dtype `np.float32` (rather than the usual default `np.float64`), the parameters passed to the likelihood function on each iteration should also be 4-byte floats.

### Actual behavior

On the _first_ call to the likelihood function, everything is as expected: the input parameter array has the same dtype as the initial parameters passed to `run_mcmc`. On the _second_ call, however, the parameter array has a different size (half the original on the zeroth axis) and the dtype is `np.float64`.

### What I have tried so far

Inspired by the test for quad precision in the repo, I tried using a `Backend(dtype=np.float32)`, but that doesn't seem to make any difference. I also looked for instances of explicit `dtype=np.float64` in the codebase, but it rarely appears, and my quick attempt at a fix in the `WalkMove` class also didn't change the observed behavior.

### Minimal example

```python
import emcee
import numpy as np

sigma = 1.

def lhood(p):
    if p.dtype != np.float32:
        print(p.dtype, p.shape)
        raise ValueError                                                                                               
    return -0.5 * np.sum(p**2, axis=1) / sigma**2

nwalkers, ndim = 16, 4
sampler = emcee.EnsembleSampler(nwalkers, ndim, lhood, vectorize=True)
p0 = np.random.randn(nwalkers, ndim).astype(np.float32)
sampler.run_mcmc(p0, 100)
```

This will print, before raising a `ValueError`: `np.float64 (8, 4)`. Same total array size, but reinterpreted as a different data type.

dfm:
Thanks! I expect that what's happening here is that all the calls to `numpy.random` are producing `float64` arrays which can be upcast to `float128`, but won't be downcast to `float32`. This means that every time a proposal is made, the coordinates will end up being cast to the higher precision. It would take some work to be explicit about the random dtype throughout the whole code base, so I'm not sure there's much I'd want to change about emcee's operation here. For your use case, would it be sufficient to just downcast the dtype at the top of `lhood` before evaluating your model on the GPU?

emprice:
[...]

PR #484, opened 2023-10-17
Move proposals should match the dtype of original input

emprice:
This PR addresses #483 for the `WalkMove` and `DESnookerMove` classes by replacing `np.empty` with `np.empty_like`, so the same datatype as the input walker positions is used in new proposals.


dfm:
Thanks!!

PR #488, opened 2023-11-02
Fix moves documentation typos

teobouvard:
This commit fixes a display issue in the class reference, as well as a typo in the same sentence.


dfm:
Thanks!

PR #506, opened 2024-03-26
Address numpy deprecation warning

OriolAbril:
I am seeing:

```
DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
```

from these lines, I think this would fix them. Not really sure how to add a test for that or if you'd prefer a diferent approach.

### DIFF
--- a/src/emcee/ensemble.py
+++ b/src/emcee/ensemble.py
@@ -489,10 +489,11 @@ class EnsembleSampler(object):
             results = list(map_func(self.log_prob_fn, p))
 
         try:
-            log_prob = np.array([float(l[0]) for l in results])
+            log_prob_elements = [l[0] for l in results]
+            log_prob = np.array([float(l.item() if hasattr(l, "item") else l) for l in log_prob_elements])
             blob = [l[1:] for l in results]
         except (IndexError, TypeError):
-            log_prob = np.array([float(l) for l in results])
+            log_prob = np.array([float(l.item() if hasattr(l, "item") else l) for l in results])
             blob = None
         else:
             # Get the blobs dtype
--- a/src/emcee/ensemble.py
+++ b/src/emcee/ensemble.py
@@ -490,10 +490,17 @@ class EnsembleSampler(object):
 
         try:
             log_prob_elements = [l[0] for l in results]
-            log_prob = np.array([float(l.item() if hasattr(l, "item") else l) for l in log_prob_elements])
+            log_prob = np.array(
+                [
+                    float(l.item() if hasattr(l, "item") else l)
+                    for l in log_prob_elements
+                ]
+            )
             blob = [l[1:] for l in results]
         except (IndexError, TypeError):
-            log_prob = np.array([float(l.item() if hasattr(l, "item") else l) for l in results])
+            log_prob = np.array(
+                [float(l.item() if hasattr(l, "item") else l) for l in results]
+            )
             blob = None
         else:
             # Get the blobs dtype
--- a/src/emcee/ensemble.py
+++ b/src/emcee/ensemble.py
@@ -490,17 +490,17 @@ class EnsembleSampler(object):
 
         try:
             log_prob_elements = [l[0] for l in results]
-            log_prob = np.array(
-                [
-                    float(l.item() if hasattr(l, "item") else l)
-                    for l in log_prob_elements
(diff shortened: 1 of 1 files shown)


dfm:
Thanks! Can you share a code snippet and all the relevant versions where you're seeing this? I can't reproduce this warning.

OriolAbril:
I think the default is to not show DeprecationWarning for users. Try setting

```python
import warnings
warnings.simplefilter('always')
```

I see it on anything with blobs. haven't checked multiple versions, I am on python 3.10, numpy 1.26.4 and 3.1.4.

Some example snippets:

```python
# single blob
sampler = emcee.EnsembleSampler(6, 1, lambda x: (-(x**2), 3))
sampler.run_mcmc(np.random.normal(size=(6, 1)), 20);
# multiple blobs
sampler = emcee.EnsembleSampler(6, 1, lambda x: (-(x**2), (np.random.normal(x), 3)))
sampler.run_mcmc(np.random.normal(size=(6, 1)), 20);
```

dfm:
The problem here is really that there is a bug in the log probability that you're specifying. The log probability should always be a scalar, but in this case it has size `ndim`. The "correct" log prob is probably something like:

```diff
- lambda x: (-(x**2), 3)
+ lambda x: (-np.sum(x**2), 3)
```

I'd actually probably rather raise an exception for non-scalar probability, rather than trying to coerce it into a scalar. I expect that this is a fairly small corner case (log probs that have shape `(1,)` instead of `()`)...

That being said, relying on numpy's deprecation for this doesn't seem like a good approach either! Would you be willing to update this PR to explicitly check that the log prob is a scalar?

OriolAbril:
Sounds good, I thought (1,) and scalars were exchangeable, I'll fix my code too. Do you have a preferred way to check something is a scalar? `np.ndim(x) == 0`?

dfm:
I guess they used to be interchangeable, but if numpy is deprecating this behavior, it might be best for emcee to follow suit!

> Do you have a preferred way to check something is a scalar? `np.ndim(x) == 0`?

This seems sensible to me. Thank you!!

andyfaff:
In scipy we come across this rather more than I'd like, so much jumping through hoops. Anyway for a multivariate function returning a scalar we deal with it [as](https://github.com/scipy/scipy/blob/2ecac3e596fdb458c85000e7707a8f5f46926621/scipy/optimize/_differentiable_functions.py#L147):

```
            fx = fun(np.copy(x), *args)
            # Make sure the function returns a true scalar
            if not np.isscalar(fx):
                try:
                    fx = np.asarray(fx).item()   # deals with 
                except (TypeError, ValueError) as e:
                    raise ValueError(
                        "The user-provided objective function "
                        "must return a scalar value."
                    ) from e
```

It should deal with `1.0, np.double(1.0), np.array(1.0), np.array([1.0])`

OriolAbril:
I understood the goal wasn't to deal with `np.array([1.0])` but to raise an error if encountered, hence the change from the inital use of `item` to ndim and explicit error raising.

dfm:
Thanks @andyfaff!! This is very useful context.

It's interesting that scipy has decided to handle the deprecation this way, and it definitely makes me less certain about how we want to navigate this. The original motivation for emcee was to be roughly compatible with scipy's `optimize` interface, so maybe it is right to match that.

Another more radical approach would be to explicitly sum the log prob result down to a scalar, but that would be a bigger change to the expected interface.

In the end I guess I don't have a strong opinion so I'm happy with either raising the error or using the scipy approach that @andyfaff shared above. What do you think @OriolAbril?

Thanks both!!

andyfaff:
I've dealt with this issue, and added unit tests as part of #510. I went with the fix that I suggested.

As mentioned above it will deal with `1.0, np.float64(1.0), np.array(1.0), np.array([1.0])`. The latter is quite common in user-land, #509.

OriolAbril:
> I've dealt with this issue, and added unit tests as part of https://github.com/dfm/emcee/pull/510. I went with the fix that I suggested.

closing this then

dfm:
Thanks, @OriolAbril! And thanks again for tracking this down. I'm happy with where this landed in the end.

Issue #515, opened 2024-04-16
New release to prepare for numpy 2


MuellerSeb:
Since the first release candidate of numpy 2 is out ([numpy 2.0.0rc1](https://pypi.org/project/numpy/2.0.0rc1)) and the ABI is stable with that, packages start building against numpy 2 (e.g. [scipy](https://github.com/scipy/scipy/blob/v1.13.0/pyproject.toml#L38)) to have a smooth transition once the release is out.

Since #510 is merge and #509 is solved, it would be good to have a release of emcee that is ready for numpy 2 so dependent packages like [GSTools](https://github.com/GeoStat-Framework/GSTools) can also start building against numpy 2 and also test it.

Could you create such a release in the near future or are there any other things blocking a new release of emcee?

ATM we have to use a hack to install the current develop version of emcee to test our package in the CI ([link](https://github.com/GeoStat-Framework/GSTools/pull/340/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R117))

Thanks!

dfm:
Done: https://pypi.org/project/emcee/3.1.5/

Please report back if there are any issues!

jwhhh:
Hi team, I haven't looked into the details yet, but `emcee v3.1.5` seems to conflict with `numpy v1.24.4` (the highest numpy version under python 3.8). I get the following error message under the mentioned version:

```console
AttributeError: module 'numpy' has no attribute 'exceptions'
```

Is there any possible fix (e.g. specifying `emcee v3.1.5` to only require `numpy >= 2`, or do I have to just wait for numpy to officially release the new major version?

andyfaff:
You can install the numpy 2.0.0 release candidate from pypi, but only for python>=3.9.

MuellerSeb:
You need at least numpy 1.25.0: https://numpy.org/doc/1.25/release/1.25.0-notes.html#numpy-now-has-an-np-exceptions-namespace

There should be a workaround to support earlier versions. Will do a PR.

MuellerSeb:
@dfm this should get another bug fix release, since emcee 3.1.5 unintentionally dropped support for a lot of numpy versions.

PR #517, opened 2024-04-19
Compat fix for older numpy versions

MuellerSeb:
Fixes: https://github.com/dfm/emcee/issues/515#issuecomment-2065800178

Before numpy 1.25, there was no `exceptions` namespace in numpy: https://numpy.org/doc/1.25/release/1.25.0-notes.html#numpy-now-has-an-np-exceptions-namespace

This PR fixes this for older numpy versions by importing `VisibleDeprecationWarning` from top level if `exceptions` is not present.

### DIFF
--- a/src/emcee/ensemble.py
+++ b/src/emcee/ensemble.py
@@ -21,6 +21,13 @@ except ImportError:
     # for py2.7, will be an Exception in 3.8
     from collections import Iterable
 
+try:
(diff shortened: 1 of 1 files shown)


dfm:
Thanks @MuellerSeb!! I've added a job to the tests workflow which should test this issue. As soon as that finishes running I'll merge.

PR #540, opened 2025-08-18
Increase sampling diversity by RedBlue one-to-one mapping

JohannesBuchner:
Currently, the stretch move resamples again. Instead of finding a one-to-one mapping of red and blue, this can reuse points, decreasing proposal diversity.

This commit favors a one-to-one mapping.

### DIFF
--- a/src/emcee/moves/stretch.py
+++ b/src/emcee/moves/stretch.py
@@ -33,14 +33,15 @@ class StretchMove(RedBlueMove):
         # deterministic matching of c to s
         # they should be the same size +- 1.
         if Nc == Ns:
-            cc = c
+            rint = Ellipsis
         elif Nc == Ns + 1:  # s is one shorter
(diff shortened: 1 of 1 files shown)


dfm (changes requested):
Thanks! Looks good but here are some high level comments:

1. I'd prefer not to change default behavior. If we add this behavior, let's put it under an option that gets passed to the move. This project is pretty much in maintenance mode and I don't think it's a great idea to dramatically change default behavior like this.

2. This won't necessarily play nicely with the `nsplits` and `randomize_split` parameters to the `RedBlueMove`. If `nsplits != 2` then the `else` will always be hit, and if `randomize_split is False` then this will deterministically reuse the same helper walker every time. Perhaps a better approach would be to shuffle `arange(Nc)` and then append `randint(Nc, size=(Ns-Nc,))` if `Nc < Ns`?

PR #545, opened 2025-09-24
Use up-to-date tqdm interface and add unit tests

vandalt:
The `tqdm.tqdm_notebook` interface causes a deprecation warning. The new way to import tqdm subodules is `from tqdm.notebook import tqdm`.

This PR changes the `pbar.py` module to use this mechanism through importlib without breaking the user-facing interface. This also enables the use of `tqdm.auto`, which is nice when emcee is used inside another package where users cannot directly set `progress`.

One thing I was wondering: would it be OK to make the default progress bar use `tqdm.auto` instead of `tqdm.tqdm`? That way the nice notebook progress bar would always be used in notebooks. If so I would update the PR with this.

Thank you!

### DIFF
--- a/src/emcee/pbar.py
+++ b/src/emcee/pbar.py
@@ -9,6 +9,7 @@ logger = logging.getLogger(__name__)
 
 try:
     import tqdm
+    import tqdm.auto
 except ImportError:
     tqdm = None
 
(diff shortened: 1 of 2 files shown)


vandalt:
Done!

dfm (approved):
Thanks!

PR #571, opened 2026-07-12
Draw an independent proposal per walker for full-covariance GaussianMove

steps-re:
When `GaussianMove` is built with a full (2D) covariance matrix, `rng.multivariate_normal(...)` was called without `size=`, so it drew a single random vector that got broadcast identically onto every walker's coordinates. That destroys inter-walker independence, which ensemble MCMC relies on. The fix adds `size=x0.shape[0]` so each walker gets its own draw.

Reproduced on current main: all 8 walkers received bit-identical proposal deltas. Existing integration tests missed it because they only check aggregate marginal statistics, not per-walker independence. Added `test_full_covariance_proposal_is_independent_per_walker`, which fails before the fix and passes after. Full suite 129/129.


dfm (approved):
Thanks!

PR #572, opened 2026-07-12
Avoid mutating sampler.pool as a side effect of pickling

steps-re:
`EnsembleSampler.__getstate__` did `d = self.__dict__` (a reference, not a copy) and then `d["pool"] = None`. Any pickle, `copy.copy`, or checkpoint of the sampler therefore permanently nulled the original object's `pool`, silently breaking parallel sampling for the rest of the run. The fix copies the dict first: `d = self.__dict__.copy()`.

Reproduced with a real `multiprocessing.Pool`: the live sampler's pool became `None` after `pickle.dumps`. Added `test_pickle_does_not_mutate_original_pool`, which fails before and passes after. Full suite 129/129.


dfm (approved):
Thanks!

